PR #44, opened 2015-07-01
Reset event store recorded events collection before triggering post commit event

dottorbabba:
When the event store triggers the post commit event all the transaction related aspect must be closed. The collection of recorded events stored during the transaction and made persistent into the event store
must be empty.
If the listener of the "post commit" event wants to start a new transaction it have not to take care of the reset of the recorded events collection. It's not its responsibility. If the collection isn't empty the events of the new transaction will be added to the old ones and sent at the new post commit event generating a loop.
If the listener is interested into what events has been committed it can look into the event itself using the `getRecordedEvents` method.


codeliner:
:+1: very good catch. I totally missed that

Issue #57, opened 2015-08-14
Move event-store related factories from proophessor to this repo


codeliner:
Since zend-servicemanager v2.6 the zend service manager implements `ContainerInterface`.
So a zend specific factory is no longer needed and we can provide invokable factories instead. This allows the usage in any environment working with a `ContainerInterface`. The only requirement is is that a application wide `config` is registered as a `service` in the container. But this is a common and useful feature anyway.

Example Factory

```php
final class EventStoreFactory
{
    public function __invoke(ContainerInterface $container)
    {
        $eventStoreConfig = $container->get(\Prooph\EventStore\Configuration\Configuration::class);

        if ($container->has(\Prooph\Common\Messaging\MessageFactory::class) {
            $eventStoreConfig->setMessageFactory(
                $container->get(\Prooph\Common\Messaging\MessageFactory::class);
            );
        }

        //get ES Adapter and other aspects ...

        return new EventStore($eventStoreConfig);
    }
}
````

Such a factory will centralize the event store initialization and increase re usability of the event-store component a lot.

prolic:
The event store factory makes use of the ZF2FeatureManager, which is clearly ZF2 dependend. Do you want to provide a ContainerInteropProxy for that reason?

prolic:
Second problem: RepositoryAbstractFactory implements Zend\ServiceManager\AbstractFactoryInterface

codeliner:
@prolic we need to align the factories. Check out the [5.0-beta branch](https://github.com/prooph/event-store/tree/5.0-beta) prooph/event-store no longer depends on ZF2 components only container-interop. So all factories should also only type-hint the interface

codeliner:
Done with #67

PR #66, opened 2015-08-20
[wip] Move event-store related factories from proophessor to this repo

prolic:
see: https://github.com/prooph/event-store/issues/57

PR #67, opened 2015-08-20
Move event-store related factories from proophessor to this repo

prolic:
- Update event store constructor
- create new factories
- update tests

see: https://github.com/prooph/event-store/issues/57


prolic:
related to: https://github.com/prooph/event-store/pull/66

Issue #72, opened 2015-08-21
Improve set up of AggregateRepository


codeliner:
- [x] ~~We should also provide a factory for the AggregateRepository.~~ -> moved to #89 

- [x] Furthermore, stream strategy should default to SingleStreamStrategy.

That means constructor arguments will change

```php
    /**
     * @param EventStore $eventStore
     * @param AggregateType $aggregateType
     * @param AggregateTranslator $aggregateTranslator
     * @param StreamStrategy|null $streamStrategy
     */
    public function __construct(
        EventStore $eventStore,
        AggregateType $aggregateType,
        AggregateTranslator $aggregateTranslator,
        StreamStrategy $streamStrategy = null
    ) {
```

prolic:
- constructor changed
- single stream strategy is now default

todo: write factory and factory test

codeliner:
@prolic I've updated the description. Would like to schedule the factory for v5.1. It is not a must have atm.
shipping 5.0 has higher priority.

PR #76, opened 2015-08-21
Change constructor signature of AggregateRepository

prolic:
see: https://github.com/prooph/event-store/issues/72


codeliner:
@prolic Good evening :smile: 
Unfortunately, I don't have much time today (have to work on a client project) but I merged this one if you want to work on the second part #72

PR #81, opened 2015-08-22
better phpunit xml file

prolic:


prolic:
I'll push this to the php-cs-fixer branch, as it depends on it.

PR #82, opened 2015-08-22
Remove adapter helper trait

codeliner:
See #64


codeliner:
ping @prolic

codeliner:
I think we can ignore the reduced coverage at the moment. I created a new issue to tackle it: #84
We should finish the other tasks first and if we'll be still below 85 % we can see how we can improve the tests.

prolic:
:+1:

PR #87, opened 2015-08-25
Avoid config conflicts with adapter factories

codeliner:
While documenting the event store factory I recognized that we have a conflict in the config structure.
The event store factory expects a `string` for `prooph.event_store.adapter` while the adapter factories expect an `array` containing a `options` key.

To avoid such conflicts this PR changes the event store factory to look for a `type` key within the `adapter` config for the service id of the adapter.


codeliner:
ping @prolic

PR #95, opened 2015-09-19
apply events late

prolic:
it applies events to aggregate roots only after they are written to the event stream

### DIFF
--- a/src/Aggregate/ConfigurableAggregateTranslator.php
+++ b/src/Aggregate/ConfigurableAggregateTranslator.php
@@ -80,7 +80,7 @@ class ConfigurableAggregateTranslator implements AggregateTranslator
         }
 
         if (null !== $applyRecordedEventsMethodsName) {
-            Assertion::minLength($applyRecordedEventsMethodsName, 1, 'Pop recorded events method name needs to be a non empty string');
+            Assertion::minLength($applyRecordedEventsMethodsName, 1, 'Apply recorded events method name needs to be a non empty string');
             $this->applyRecordedEventsMethodName = $applyRecordedEventsMethodsName;
         }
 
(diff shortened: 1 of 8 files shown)


src/Aggregate/ConfigurableAggregateTranslator.php
```diff
@@ -72,6 +79,11 @@ public function __construct(
             $this->popRecordedEventsMethodName = $popRecordedEventsMethodName;
         }
 
+        if (null !== $applyRecordedEventsMethodsName) {
+            Assertion::minLength($applyRecordedEventsMethodsName, 1, 'Pop recorded events method name needs to be a non empty string');
```
codeliner: s/Pop recorded/Apply recorded

src/Aggregate/ConfigurableAggregateTranslator.php
```diff
@@ -211,4 +223,37 @@ public function extractPendingStreamEvents($eventSourcedAggregateRoot)
...
+
+        foreach ($events as $event) {
+            if (!$event instanceof Message) {
+                throw new AggregateTranslationFailedException(sprintf(
+                    'A recorded event of the aggregate root %s has the wrong type. Expected Prooph\Common\Messaging\Message. Got %s',
+                    get_class($eventSourcedAggregateRoot),
+                    is_object($event)? get_class($event) : gettype($event)
+                ));
+            }
+            $eventSourcedAggregateRoot->{$this->applyRecordedEventsMethodName}($event);
```
codeliner: Before applying the event we need to check if a `messageToEventCallback` was injected in the translator. If so the callback needs to be invoked first.

codeliner:
@prolic I love this change. It is the best way to solve the consistency problem :+1: It is a BC break but I cannot think of better solution without breaking BC.
Two minor issues (see inline comments) and we need to align the docs. If you want, you can create a follow up ticket for documentation and I take it over.

prolic:
done.

Issue #55, opened 2015-08-14
Remove nested transaction support


codeliner:
Based on this conversation prooph/event-store-mongodb-adapter#5

PR #97, opened 2015-09-24
remove nested transaction support

prolic:
see: https://github.com/prooph/event-store/issues/55


codeliner:
:+1: Just an addition. As the deprecated error is removed we can enable phpunit convertErrorToException again: https://github.com/prooph/event-store/blob/master/phpunit.xml.dist#L6

prolic:
give me a sec...

prolic:
done.

PR #101, opened 2015-09-28
Adjust AggregateTranslator contract

codeliner:
The ConfigurableAggregateTranslator caused trouble when reconstituting an aggregate in combination with a custom `messageToEventCallback`. In this case the translator used `array_map` but got an `\Iterator $historyEvents`.

This PR changes the `AggregateTranslator::reconstituteAggregateFromHistory` contract to require a `\Iterator $historyEvents` parameter.

Also a new `MapIterator` is introduced to replace the `array_map` functionality.

Additionally I've renamed `AggregateRepository::__invoke` to `AggregateRepository::addPendingEventsToStream` to unify the listener methods and improve readability of the code.

### DIFF
--- a/src/Aggregate/AggregateRepository.php
+++ b/src/Aggregate/AggregateRepository.php
@@ -69,7 +69,7 @@ class AggregateRepository
         StreamStrategy $streamStrategy = null
     ) {
         $this->eventStore = $eventStore;
-        $this->eventStore->getActionEventEmitter()->attachListener('commit.pre', $this);
+        $this->eventStore->getActionEventEmitter()->attachListener('commit.pre', [$this, 'addPendingEventsToStream']);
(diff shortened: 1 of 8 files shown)


codeliner:
ping @prolic

prolic:
nice! Can you add a test case that fails before the change? :)

codeliner:
@prolic [this test case](https://github.com/prooph/event-store/blob/develop/tests/Aggregate/ConfigurableAggregateTranslatorTest.php#L313) fails when you pass an \ArrayIterator as $historyEvents instead of an array. That is why I also forces the \Iterator now in the AggregateTranslator interface. The test was green because it only tested the case when $historyEvents is an array but with the "EventStream as iterator" changes this will no longer be the real case.

prolic:
:+1:

codeliner:
thanks :smile:

PR #102, opened 2015-09-28
Handle pending events of new aggregates correctly

codeliner:
While working on a new version for prooph/event-sourcing (that respects the changes made for prooph/event-store v6) I recognized that the "apply events late" feature does not work for new aggregate roots added to the repository. I was wondering why event store tests were green.
[This](https://github.com/prooph/event-store/blob/master/tests/Mock/User.php#L122) is the answer.
The mock used to test the aggregate repository was not aligned and continued to apply events directly.

This PR changes the internals of the `AggregateRepository` again to also apply events late for newly added aggregate roots. Unfortunately, this was not a straightforward task. Two problems caused a hard time. 

1. At the time when the aggregate root is added to the repository we cannot extract the aggregate id because events are not applied yet.  I've solved the problem by creating a temporary copy of the aggregate root which I can pass to the aggregate translator to get the aggregate id :P

2. Pending events for new aggregate roots need to be stored in the cache, too. So that the events are also applied in the `EventStore.commit.post` listener of the `AggregateRepository`. So fare no problem but `AggregateRepository::addPendingEventsToStream` did override the pending events for new aggregates. 

Long Story, short. I've changed the way of caching pending events to have them next to the aggregate root and have a better control over adding or clearing the events cache.

### DIFF
--- a/src/Aggregate/AggregateRepository.php
+++ b/src/Aggregate/AggregateRepository.php
@@ -52,11 +52,6 @@ class AggregateRepository
     protected $aggregateType;
 
     /**
-     * @var array
-     */
-    protected $pendingStreamEvents = [];
-
-    /**
      * @param EventStore $eventStore
      * @param AggregateType $aggregateType
      * @param AggregateTranslator $aggregateTranslator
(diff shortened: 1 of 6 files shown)


codeliner:
ping @prolic

prolic:
Really good catch! Thx :+1:

src/Aggregate/AggregateRepository.php
```diff
@@ -142,20 +144,30 @@ public function addAggregateRoot($eventSourcedAggregateRoot)
...
+        $domainEvents = $this->aggregateTranslator->extractPendingStreamEvents($eventSourcedAggregateRoot);
 
-        $domainEvents = new ArrayIterator(
-            $this->aggregateTranslator->extractPendingStreamEvents($eventSourcedAggregateRoot)
+        //We make a copy of the aggregate root to be able to extract the aggregate id
+        //without the need to apply the pending domain events to the real aggregate instance
+        //(this is done in the method applyPendingStreamEvents after EventStore.commit)
+        $aggregateRootCopy = $this->aggregateTranslator->reconstituteAggregateFromHistory(
+            $this->aggregateType,
+            new ArrayIterator($domainEvents)
```
prolic: it should be enough to apply only the first event, no reconstruction of the complete history is needed. Would you like to submit a second PR?

codeliner: good point, but at this time the history should only contain 1 event ;) If the aggregate recorded more than one event the others are maybe required to set the aggregate id correctly (really strange edge-case IMO, but we can't know it)

codeliner:
thx for fast CR and merge. Now I can finish the changes for prooph/event-sourcing

Issue #13, opened 2014-08-16
Add replay functionality


codeliner:
Maybe #37 + some documentation about how to use the event store to replay read models is enough.

We need to think about the following questions:
- [ ] List of common replay scenarios
- [ ] Is it possible to replay all events of all streams in a single operation and does it make sense?
- [ ] Is it ok to replay streams one after another?
- [ ] Are the methods `load` and `loadEventsByMetadataFrom` enough to handle common replay scenarios?

codeliner:
Maybe #37 + some documentation about how to use the event store to replay read models is enough.

We need to think about the following questions:
- [ ] List of common replay scenarios
- [ ] Is it possible to replay all events of all streams in a single operation and does it make sense?
- [ ] Is it ok to replay streams one after another?
- [ ] Are the methods `load` and `loadEventsByMetadataFrom` enough to handle common replay scenarios?

PR #104, opened 2015-10-06
Replay functionality

prolic:
solves: https://github.com/prooph/event-store/issues/13

### DIFF
--- a/src/Adapter/InMemoryAdapter.php
+++ b/src/Adapter/InMemoryAdapter.php
@@ -137,7 +137,7 @@ class InMemoryAdapter implements Adapter
         foreach ($this->streams[$streamName->toString()] as $index => $streamEvent) {
             if (null === $since && $this->matchMetadataWith($streamEvent, $metadata)) {
                 $streamEvents[] = $streamEvent;
-            } elseif((float) $streamEvent->createdAt()->format('U.u') >= (float) $since->format('U.u')
+            } elseif ((float) $streamEvent->createdAt()->format('U.u') >= (float) $since->format('U.u')
                 && $this->matchMetadataWith($streamEvent, $metadata)
             ) {
                 $streamEvents[] = $streamEvent;
--- a/src/Adapter/InMemoryAdapter.php
+++ b/src/Adapter/InMemoryAdapter.php
@@ -13,6 +13,7 @@ namespace Prooph\EventStore\Adapter;
 
 use AppendIterator;
 use ArrayIterator;
+use DateTimeInterface;
 use Iterator;
 use Prooph\Common\Messaging\Message;
 use Prooph\EventStore\Exception\StreamNotFoundException;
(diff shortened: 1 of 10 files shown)


src/Adapter/InMemoryAdapter.php
```diff
@@ -120,6 +121,33 @@ public function loadEventsByMetadataFrom(StreamName $streamName, array $metadata
...
+        if (! isset($this->streams[$streamName->toString()])) {
+            return new ArrayIterator();
+        }
+
+        $streamEvents = [];
+
+        foreach ($this->streams[$streamName->toString()] as $index => $streamEvent) {
+            if (null === $since && $this->matchMetadataWith($streamEvent, $metadata)) {
+                $streamEvents[] = $streamEvent;
+            } elseif((float) $streamEvent->createdAt()->format('U.u') >= (float) $since->format('U.u')
```
codeliner: why typecasting here? you can use \DateTime(Immutable) with normal comparison

src/EventStore.php
```diff
@@ -254,6 +257,39 @@ public function loadEventsByMetadataFrom(StreamName $streamName, array $metadata
...
+        $iterators = [];
+        foreach ($streamNames as $key => $streamName) {
+            $iterators[] = $this->adapter->replay($streamName, $since, $metadatas[$key]);
+        }
+
+        return new CompositeIterator($iterators, function(Message $message1 = null, Message $message2) {
+            if (null === $message1) {
+                return true;
+            }
+            return (float) $message1->createdAt()->format('U.u') > (float) $message2->createdAt()->format('U.u');
```
codeliner: same as above. I guess typecast is not needed

codeliner: Damn, or do we need the type cast because PHP does not compare the microseconds correctly?

src/Adapter/InMemoryAdapter.php
codeliner: Damn, or do we need the type cast because PHP does not compare the microseconds correctly?

src/EventStore.php
codeliner: Depending on your answer above type cast is maybe not needed.

src/Util/CompositeIterator.php
```diff
@@ -0,0 +1,113 @@
...
+            throw new \InvalidArgumentException();
+        }
+
+        foreach ($iterators as $iterator) {
+            if ($iterator instanceof IteratorAggregate) {
+                $iterator = $iterator->getIterator();
+            }
+
+            if (! $iterator instanceof Iterator) {
+                throw new \InvalidArgumentException();
```
codeliner: Don't forget to add an exception message here

tests/Adapter/InMemoryAdapterTest.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Prooph\EventStoreTest\Adapter;
 
+use Guzzle\Stream\Stream;
```
codeliner: Guzzle stream?

codeliner:
Looks great :+1: . Nice, that the replay method for adapters is simple and that the heavy work is done by the event store and the composite filter.

PR #106, opened 2015-10-08
Introduce identity map as first-class object

codeliner:
This PR includes several changes:

- IdentityMap becomes a first-class object 
- `AggregateRepository::__construct` got a new (optional) parameter `IdentityMap $identityMap = null`
    - This allows injection of a custom identity map which is capable of caching and providing aggregate roots
    - This allows the user to use for example memcached to keep aggregates in memory between requests 
    - and/or inject an IdentityMap which can load aggregate roots from snapshots
- The AggregateRepository now asserts that an added aggregate root has the correct aggregate type
    - **BC break!**: In the past it was possible to inherit from a base class - let's say `User` and handle `SalesManager`, `Bookkeeper`, etc. with the same UserRepository. This is still possible but the base class needs to implement [AggregateTypeProvider](https://github.com/prooph/event-store/blob/master/src/Aggregate/AggregateTypeProvider.php) otherwise `UserRepository::addAggregateRoot($salesManager)` will fail.

### DIFF
--- a/src/Aggregate/AggregateRepository.php
+++ b/src/Aggregate/AggregateRepository.php
@@ -13,7 +13,6 @@ namespace Prooph\EventStore\Aggregate;
 
 use ArrayIterator;
 use Assert\Assertion;
-use Prooph\EventStore\Aggregate\Exception\AggregateTypeException;
 use Prooph\EventStore\EventStore;
 use Prooph\EventStore\Stream\SingleStreamStrategy;
 use Prooph\EventStore\Stream\StreamStrategy;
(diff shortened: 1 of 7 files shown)


codeliner:
Damn, I don't know what causes coverage to decrease. @prolic Do you have an idea?

prolic:
About coverage:

InMemoryIdentityMap.php::L76

    if (! $this->has($aggregateType, $aggregateId)) {
        return; // not tested
    }

Will do further review later.

codeliner:
thx @prolic I overlooked it.

prolic:
perfect

codeliner:
:fireworks:

PR #111, opened 2015-10-08
remove has-method from identity map

prolic:
The has method must be removed from the identity map.

1) It's not needed.
2) When using non-in-memory-identity-maps this will lead to 2 queries, but one is sufficient.


prolic:
Added markDirty method to identity map.

codeliner:
very good improvements :+1:

Issue #26, opened 2014-09-02
Document how a MessageFacory can be used to load older versions of events 


codeliner:
~~see axon framework for inspiration:  http://www.axonframework.org/docs/2.0/repositories-and-event-stores.html#event-upcasting~~

Update: Who needs the axon framework :smile:

codeliner:
@prolic The more I think about it the more I think you are right with your statement that we've solved the problem with the `MessageFactory` approach. Nice bonus. I keep the ticket open because we should at least document it in a cookbook.

PR #120, opened 2015-10-14
Upcasting the right way

codeliner:
See #26


codeliner:
and the next one @prolic

PR #122, opened 2015-10-15
Allow metadatas to be empty

codeliner:
While writing docs I recognized that the signature of `EventStore::replay` is a bit confusing. The second parameter `$since` is optional but the third one needs to be given again. My first idea was to flip the parameters but then I thought it would also be a good idea to turn `$metadatas` into an optional parameter, too. @prolic Let me know what you think about the change.

### DIFF
--- a/src/EventStore.php
+++ b/src/EventStore.php
@@ -271,7 +271,7 @@ class EventStore
 
         if (null === $metadatas) {
             $metadatas = [];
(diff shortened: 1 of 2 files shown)


prolic:
Valid point, but travis does not agree :-)

codeliner:
travis should go home and his friend CS, too

Issue #116, opened 2015-10-11
Make use of interop config


prolic:

codeliner:
especially for #89 we should use interop-config

PR #128, opened 2015-10-21
Make use of interop config for event store factory

prolic:
see: https://github.com/prooph/event-store/issues/116


prolic:
Closed, new here: https://github.com/prooph/event-store/pull/130

damn travis

PR #134, opened 2015-11-02
add aggregate type and id as metadata to all stream strategies

prolic:


codeliner:
doctrine adapter is aligned in https://github.com/prooph/event-store-doctrine-adapter/pull/26

PR #135, opened 2015-11-02
extractVersion in aggregate translator

prolic:


src/Aggregate/ConfigurableAggregateTranslator.php
```diff
@@ -64,6 +70,7 @@ class ConfigurableAggregateTranslator implements AggregateTranslator
      * @param null|callable $messageToEventCallback
      */
     public function __construct(
+        $versionMethodName = null,
```
codeliner: Please move to second parameter position as the identifierMethodName is the one which is customized most often

src/Aggregate/ConfigurableAggregateTranslator.php
```diff
@@ -71,6 +78,11 @@ public function __construct(
         $eventToMessageCallback = null,
         $messageToEventCallback = null)
     {
+        if (null !== $versionMethodName) {
+            Assertion::minLength($versionMethodName, 1, 'Version method name needs to e a non empty string');
```
codeliner: should be: needs to **b**e a non empty string

codeliner:
Documentation required. Handled in issue: https://github.com/prooph/event-store/issues/136

codeliner:
@prolic CS issue + minor CR issues

PR #147, opened 2015-11-21
Rewind stream events when creating a stream

codeliner:
Small bugfix for InMemoryAdapter. When creating a new stream the recorded stream events iterator must be rewind as otherwise a later call to `$streamEvents->valid()` will be false


codeliner:
Oh, please don't merge yet! I'll add a rewind to loadstream, too

codeliner:
ok, done

PR #151, opened 2016-02-03
Upgrade depencency to ramsey/uuid 3.x

robertlemke:
This change bases the existing code on the 3.x version of ramsey/uuid (previously called rhumsaa/uuid) in order to avoid conflicts with third-party libraries which require version 3.0 or higher.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     ],
     "require": {
         "php": "~5.5|~7.0",
-        "ramsey/uuid" : "~2.8",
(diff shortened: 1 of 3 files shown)


codeliner:
Same like https://github.com/prooph/common/pull/52

prolic:
Btw: Tests are failing.

codeliner:
Closing for now, as with #166 uuid becomes a dev dependency only

PR #157, opened 2016-02-26
Metadata enricher

MattKetmo:
Hello,

Here is a small plugin I wrote for a project to easily add custom metadata to all events. The code is inspired from [event-store-bus-bridge/src/TransactionManager.php#L54](https://github.com/prooph/event-store-bus-bridge/blob/master/src/TransactionManager.php#L54).

In my case I use it to add information about the user who caused the events.

The PR is missing the tests and some phpdoc but I prefer to ask before, to know if you want the feature in this package or prefer having it elsewhere (or maybe you don't want it). 

I also consider to configure it in the Container via something like:

```php
return [
    'prooph' => [
        'event_store' => [
            'metadata_enrichers' => [
                UserMetadataEnricher::class,
                OtherMetadataEnricher::class,
            ],
    // ...
```

(Btw this is also something [Broadway brings natively](https://github.com/qandidate-labs/broadway/blob/master/src/Broadway/EventSourcing/MetadataEnrichment) ;))

### DIFF
--- a/src/Metadata/MetadataEnricher.php
+++ b/src/Metadata/MetadataEnricher.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the prooph/event-store package.
+ * (c) 2014 - 2016 prooph software GmbH <contact@prooph.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Prooph\EventStore\Metadata;
+
+use Prooph\Common\Messaging\Message;
+
+interface MetadataEnricher
+{
+    /**
+     * Return the given message with added metadata.
+     *
+     * @param Message $message
+     *
+     * @return Message
+     */
+    public function enrich(Message $message);
+}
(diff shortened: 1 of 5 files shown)


src/Metadata/MetadataEnricherAggregation.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Prooph\EventStore\Metadata;
+
+use Prooph\Common\Messaging\Message;
+
+final class MetadataEnricherAggregation implements MetadataEnricher
```
prolic: i would suggest the name MetadataEnricherAggregate

prolic:
I like it. Let's add some phpdoc and test-coverage back to 100% :-)
@codeliner additional thoughts?

codeliner:
I like it, too. It reminds me on refactoring the TransactionManager so that causation_id and causation_name are no longer added by the TransactionManager. If the MetadataEnricherPlugin is available we can change the behavior of TransactionManager with the next major release.

prolic:
@MattKetmo can you submit the pr to the develop branch, please?

MattKetmo:
ok, I will do it tomorrow

MattKetmo:
PR submitted in develop branch with tests and phpdoc #162.

PR #159, opened 2016-02-27
Set recordedEvent back to ArrayIterator when rolling back transaction

prolic:
... also updates coveralls to ^1.0


prolic:
@codeliner this! :)

prolic:
I hope it's okay to do all in one PR, if neccessary I can split into multiples, too.

codeliner:
that's a good question. Would be good to have the bugfix in a separate PR so that we can link to it in the changelog/release note. I'm off again for 2 hours and merge it later.

prolic:
see: https://github.com/prooph/event-store/pull/160 & https://github.com/prooph/event-store/pull/161

PR #162, opened 2016-02-27
Metadata enricher

MattKetmo:
Supersede #157

### DIFF
--- a/src/Metadata/MetadataEnricherAggregate.php
+++ b/src/Metadata/MetadataEnricherAggregate.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the prooph/event-store package.
+ * (c) 2014 - 2016 prooph software GmbH <contact@prooph.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Prooph\EventStore\Metadata;
+
+use Assert\Assertion;
+use Prooph\Common\Messaging\Message;
+
+final class MetadataEnricherAggregate implements MetadataEnricher
+{
+    /**
+     * @var MetadataEnricher[]
+     */
+    private $metadataEnrichers;
+
+    /**
+     * @param MetadataEnricher[] $metadataEnrichers
+     */
+    public function __construct(array $metadataEnrichers)
+    {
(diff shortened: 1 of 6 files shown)


src/Metadata/MetadataEnricherAggregate.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace Prooph\EventStore\Metadata;
+
+use Prooph\Common\Messaging\Message;
+
+final class MetadataEnricherAggregate implements MetadataEnricher
+{
+    /**
+     * @var array
```
codeliner: type hint MetadataEnricher[]

src/Metadata/MetadataEnricherAggregate.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    private $metadataEnrichers;
+
+    /**
+     * @param array $metadataEnrichers
```
codeliner: type hint MetadataEnricher[]

src/Metadata/MetadataEnricherAggregate.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function __construct(array $metadataEnrichers)
+    {
+        $this->metadataEnrichers = $metadataEnrichers;
```
codeliner: Assert::allInstanceOf(...)

codeliner:
@MattKetmo Looks great! Just two minor remarks and we should add a note in the docs. I'd say a short section at the end of the [event_store.md](https://github.com/prooph/event-store/blob/master/docs/event_store.md) file is enough. We've defined two ways for adding docs, either with the PR or by creating a doc issue and link it with the PR. Goal is to have both ready before the feature is merged into master.

MattKetmo:
Type hint fixed + "*Metadata enricher*" section [added to the doc](https://github.com/MattKetmo/event-store/blob/metadata-enricher/docs/event_store.md#metadata-enricher) with an example of usage.

I will later submit another PR to integrate this with the ContainerInterop as suggested in previous PR. Something like:

```php
return [
    'prooph' => [
        'event_store' => [
            'metadata_enrichers' => [
                UserMetadataEnricher::class,
                OtherMetadataEnricher::class,
            ],
    // ...
```

PR #164, opened 2016-02-27
Setup metadata_enrichers option in event store factory

MattKetmo:
As discussed in #162.


prolic:
:+1: great! Thanks for the PR

PR #166, opened 2016-04-20
Move ramsey/uuid to require-dev

sbacelic:
As ramsey/uuid is currently being used only in tests it can be moved to require-dev

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,13 +23,13 @@
     ],
     "require": {
         "php": "~5.5|~7.0",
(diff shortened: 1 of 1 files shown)


sandrokeil:
Good catch! :+1:  This fixes #151 too (not for development). @codeliner What do you think, ca we bring a patch release here, because it's not a dependency of the library itself.

codeliner:
thx, patch release is on the way

PR #168, opened 2016-05-24
Fix/example array to iterator

anthonysterling:
Prooph\EventStore\EventStore::appendTo and Prooph\EventStore\Stream\Stream::__construct
expect an Iterator, this wraps the arrays being passed to them in ArrayIterator.


prolic:
Thank you!

PR #176, opened 2016-07-14
Foreach loop can be optimised into 1 internal function call.

bweston92:


sandrokeil:
I guess we need a benchmark suite ;-)

PR #181, opened 2016-09-05
Revise files header license - add docheader

prolic:


malukenho:
Still need some CS fixes :D

PR #182, opened 2016-09-08
Introduce EventStore#transactional(callable $callable) API

malukenho:
/cc @prolic

### DIFF
--- a/src/EventStore.php
+++ b/src/EventStore.php
@@ -300,7 +300,7 @@ class EventStore
      *
      * @return mixed
      */
-    public function transaction(callable $callable)
+    public function transactional(callable $callable)
     {
         $this->beginTransaction();
 
--- a/src/EventStore.php
+++ b/src/EventStore.php
@@ -298,17 +298,24 @@ class EventStore
     /**
      * @param callable $callable
      *
+     * @throws \Exception
+     *
      * @return mixed
      */
     public function transactional(callable $callable)
     {
         $this->beginTransaction();
 
-        $result = $callable($this);
+        try {
+            $result = $callable($this);
+            $this->commit();
+        } catch (\Exception $e) {
+            $this->rollback();
 
-        $this->commit();
+            throw $e;
+        }
 
-        return $result;
+        return $result ?: true;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/EventStore.php
```diff
@@ -296,6 +296,22 @@ public function replay(array $streamNames, DateTimeInterface $since = null, arra
     }
 
     /**
+     * @param callable $callable
+     *
+     * @return mixed
+     */
+    public function transaction(callable $callable)
```
prolic: rename method to _transactional_

src/EventStore.php
```diff
@@ -296,6 +296,22 @@ public function replay(array $streamNames, DateTimeInterface $since = null, arra
...
+    {
+        $this->beginTransaction();
+
+        $result = $callable($this);
```
prolic: try and catch, when exception rollback and throw again

src/EventStore.php
```diff
@@ -296,6 +296,22 @@ public function replay(array $streamNames, DateTimeInterface $since = null, arra
...
+
+        $this->commit();
+
+        return $result;
```
prolic: return $result ?: true;

src/EventStore.php
```diff
@@ -296,6 +296,29 @@ public function replay(array $streamNames, DateTimeInterface $since = null, arra
...
+        try {
+            $result = $callable($this);
+            $this->commit();
+        } catch (\Exception $e) {
+            $this->rollback();
+
+            throw $e;
+        }
+
+        return $result ?: true;
```
prolic: test for return true missing

prolic:
test for exception and rollback missing

tests/Mock/AdapterCanHandlerMock.php
```diff
@@ -0,0 +1,44 @@
...
+use Prooph\EventStore\Adapter\Feature\CanHandleTransaction;
+use Prooph\EventStore\Adapter\InMemoryAdapter;
+
+/**
+ * Class AdapterCanHandlerMock
+ *
+ * @package ProophTest\EventStore
+ * @author Jefersson Nathan <malukenho@phpse.net>
+ */
+final class AdapterCanHandlerMock extends InMemoryAdapter implements CanHandleTransaction
```
prolic: I would suggest renaming that class to TransactionalInMemoryAdapterMock, so the naming is a little more clear.

Issue #198, opened 2016-10-25
RFC load max events and direction


prolic:
The current interface is:

`public function load(StreamName $streamName, int $minVersion = null): Stream`

I want to change to:

`public function load(StreamName $streamName, int $minVersion = null, int $maxVersion = null, bool $forward = true): Stream`

to allow loading events in a given range only (useful for http api f.e., see: https://github.com/prooph/event-store/issues/194) and additionally you can load the events in reverse order.

Similar to this `loadEvents` needs to be updated, too.

basz:
What I find confusing....
Start with a min of 10. I want my set 10 not to be bigger then 10 events. Should max be 20 or 19? Do both numbers include themselves or only one of them does?

I argue the later and as such the variablenames can express that clearerly:

fromVersion, upToVersion

prolic:
:+1: for fromVersion, upToVersion

prolic:
Another thing to consider:
`version` is a term useful for event sourcing. The EventStore is capable of storing all kinds of event streams, also non-event-sourced. Thefore I like to suggest to rename `version` to `no` in the event stream. The `version` for event sourcing can be equal the `no` in the stream for one stream per aggregate usage, otherwise the `version` should be part of the metadata.

prolic:
Also: shouldn't we count the events from `0` instead from `1`?

basz:
instead of 'no' `ordinality`? (http://www.thefreedictionary.com/Ordinality)

a zero based index does not fly here. A 0'th event doesn't makes sense, only a 1st, 2nd, etc.

codeliner:
if we rename from `version` to `no` we need a migration path!

codeliner:
except the rename i like the RFC a lot :+1:

oqq:
How about to request an amount of events?
fromVersion, amountOfEvents

prolic:
@oqq my explanation in code:

```
$minVersion = 12;
$amountOfEvents = 100;
$maxVersion = $minVersion + $amountOfEvents;
```

easy?

oqq:
Of course!
But the question is, which version is way to go to simplify the api.

e.g. using a paginator you would simple call "give me the next 10 elements" instead of give me all elements from x to x+=10.

minVersion and maxVersion has to be calculated, instead of only minVersion.

prolic:
@oqq

Agree!
Let's change this to:

public function load(StreamName $streamName, int $fromNumber = null, int count = null, bool $forward = true): Stream

[...]

Issue #190, opened 2016-10-23
RFC Remove Adapter::replay method


prolic:
I am thinking about removing the `replay` method on the Adapter interface (as well as the related `CompositeIterator`).

Imho this makes not much sense and using `CompositeIterator` with millions of streams will be a big performance problem.

Therefore I suggest using another replay strategy based on the existing `load` method on the adapter interface. This can be added to the event-store-bus-bridge repository as new feature.

thoughts @codeliner ?

codeliner:
not sure... With the `load` method you can only load the history of a single stream. Replay joins events of different streams and puts them in chronological order. How do you want to achieve the same with the `load` method?

I agree, that replay becomes a problem when you have many streams. You definitely need another strategy then but for simpler use cases it is a good option and it is already there. So why remove it. Or can we move it to its own package so that not every adapter is forced to implement the replay method?

And why `event-store-bus-bridge`?

prolic:
Sorry I should have provided more information in the description.

First of all the `replay` method suggests, that this is the right way to do it as it's part of the adapter interface - but as I said this really leads to serious problems with a lot of streams (especially with one stream per aggregate).

Here my `new` example on how to make replays:

The order is very important (sometimes) and sometimes it's not. Here's why....

Let's make replay by example: I want to replay all user-streams and all friendship-streams (user-to-user-relations). This would look like this:
```
foreach ($db->fromCategory('user') as $userStream) {
    $this->replay($userStream);
}
foreach ($db->fromCategory('friends') as $friendStream) {
      $this->replay($friendStream);
}
```

So replaying one user after the other really does the job. From the standpoint of a given user, he's a single aggregate and independend of all other user-aggregates - therefore I don't need to replay them in chronolocial order, like User1-Event1 - User2-Event1 - User2-Event2 - User1-Event2 and so on - It's really okay to do it like this: User1-Event1 - User1-Event2 and then User2-Event1 - User2-Event2 - it's only important to keep the event order per stream - not between streams.

For the relations now: After I replayed all users, I can safely replay all friendship relations - it could happen, that a user is not existent anymore in the readmodel (because there was a deleted-event) - the projection needs to handle this case and reject the made-friend-with event in this case.

Note: This is also the way how replaying stream-projections in Greg's EventStore works - I tested this myself and it works without problems.

Last point:
The reason why I suggested event-store-bus-bridge (It doesn't has to be!):
[...]

Issue #196, opened 2016-10-24
RFC sha1 on stream names


prolic:
There are quite some restrictions on stream names and those restrictions are changing between different event-store-adapters, like f.e. forbidden characters in mysql vs forbidden characters in mongodb). Also the given maximum string length for a collection name in mongodb or a mysql-table name are a problem.

Therefore I like to propose to sha1 all stream names and create a table / collection where a mapping from hash to real stream name is put (so you can see what actual stream name are availalble).

codeliner:
a bit more work when working directly on database level (via mongo shell f.e.) but reasoning is valid so :+1:

oqq:
I would prefer to provide an service interface for slugify a stream name.
The default implementation could be to replace all non word characters or similar to current implementation in event store adapters.

Users could be interested to find stream tables with readable names instead of hashed ones.

prolic:
It's based on a strategy, see: https://github.com/prooph/event-store-pdo-adapter/blob/master/src/TableNameGeneratorStrategy.php

PR #200, opened 2016-10-25
RFC load max events and direction & RFC remove Adapter::replay

prolic:
Implementation for:

RFC load max events and direction https://github.com/prooph/event-store/issues/198

RFC Remove Adapter::replay method https://github.com/prooph/event-store/issues/190


prolic:
@oqq wanna review, too?

src/Adapter/Adapter.php
```diff
@@ -35,9 +34,29 @@ public function create(Stream $stream): void;
...
     public function appendTo(StreamName $streamName, Iterator $domainEvents): void;
 
-    public function load(StreamName $streamName, int $minVersion = null): ?Stream;
-
-    public function loadEvents(StreamName $streamName, array $metadata = [], int $minVersion = null): Iterator;
-
-    public function replay(StreamName $streamName, DateTimeInterface $since = null, array $metadata = []): Iterator;
+    public function load(
+        StreamName $streamName,
+        int $fromNumber = 0,
```
oqq: Is there a chance to get an event with version '0', or should it '1'?
Perhaps it is just irrelevant and '0' could stay here. Not sure..

src/Adapter/InMemoryAdapter.php
```diff
@@ -62,34 +57,74 @@ public function appendTo(StreamName $streamName, Iterator $domainEvents): void
...
+                $filteredEvents[] = $streamEvent;
+            }
+        }
 
-            foreach ($streamEvents as $streamEvent) {
-                if ($streamEvent->version() >= $minVersion) {
-                    $filteredEvents[] = $streamEvent;
+        if (null !== $count) {
+            foreach ($filteredEvents as $key => $streamEvent) {
+                if ($streamEvent->version() > ($fromNumber + $count - 1)) {
```
oqq: Is a stream always ordered by version? Maybe we could 'break' the previous loop early instead of set/unset.

src/Adapter/InMemoryAdapter.php
```diff
@@ -98,37 +133,47 @@ public function loadEvents(StreamName $streamName, array $metadata = [], int $mi
 
         foreach ($this->streams[$streamName->toString()] as $index => $streamEvent) {
             if ($this->matchMetadataWith($streamEvent, $metadata)) {
-                if (null === $minVersion || $streamEvent->version() >= $minVersion) {
-                    $streamEvents[] = $streamEvent;
+                if ($streamEvent->version() >= $fromNumber) {
+                    $streamEvents[$index] = $streamEvent;
+                }
+                if (null !== $count && $streamEvent->version() > ($fromNumber + $count - 1)) {
```
oqq: Maybe a 'continue' before 'push' is better than 'push' and 'unset'.

src/EventStore.php
```diff
@@ -388,4 +352,108 @@ public function getActionEventEmitter(): ActionEventEmitter
     {
         return $this->actionEventEmitter;
     }
+
+    private function preLoad(string $action, StreamName $streamName, int $fromNumber, int $count = null)
```
oqq: missing return type hint

src/EventStore.php
```diff
@@ -388,4 +352,108 @@ public function getActionEventEmitter(): ActionEventEmitter
...
+        $this->getActionEventEmitter()->dispatch($event);
+
+        if ($event->propagationIsStopped()) {
+            throw StreamNotFoundException::with($streamName);
+        }
+
+        return $event->getParam('stream');
+    }
+
+    private function preloadEventsByMetadataFrom(
```
oqq: missing return type hint

src/Adapter/InMemoryAdapter.php
```diff
@@ -62,34 +57,74 @@ public function appendTo(StreamName $streamName, Iterator $domainEvents): void
         $this->streams[$streamName->toString()] = $appendIterator;
     }
 
-    public function load(StreamName $streamName, int $minVersion = null): ?Stream
-    {
+    public function load(
+        StreamName $streamName,
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

see: https://github.com/prooph/event-store/issues/198#issuecomment-255976361

src/Adapter/InMemoryAdapter.php
```diff
@@ -62,34 +57,74 @@ public function appendTo(StreamName $streamName, Iterator $domainEvents): void
...
             }
+        }
+
+        return new Stream($streamName, new \ArrayIterator($filteredEvents));
+    }
 
-            return new Stream($streamName, new \ArrayIterator($filteredEvents));
+    public function loadReverse(
+        StreamName $streamName,
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

src/Adapter/InMemoryAdapter.php
```diff
@@ -62,34 +57,74 @@ public function appendTo(StreamName $streamName, Iterator $domainEvents): void
...
+
+        return new Stream($streamName, new \ArrayIterator($filteredEvents));
     }
 
-    public function loadEvents(StreamName $streamName, array $metadata = [], int $minVersion = null): Iterator
-    {
+    public function loadEvents(
+        StreamName $streamName,
+        array $metadata = [],
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

src/Adapter/InMemoryAdapter.php
```diff
@@ -98,37 +133,47 @@ public function loadEvents(StreamName $streamName, array $metadata = [], int $mi
...
 
         return new ArrayIterator($streamEvents);
     }
 
-    public function replay(StreamName $streamName, DateTimeInterface $since = null, array $metadata = []): Iterator
-    {
+    public function loadEventsReverse(
+        StreamName $streamName,
+        array $metadata = [],
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

src/EventStore.php
```diff
@@ -144,131 +142,97 @@ public function appendTo(StreamName $streamName, Iterator $streamEvents): void
...
-
-        if ($event->propagationIsStopped()) {
-            $stream = $event->getParam('stream', false);
-
-            if ($stream instanceof Stream && $stream->streamName()->toString() == $streamName->toString()) {
-                return $stream;
-            }
+    public function load(
+        StreamName $streamName,
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

src/EventStore.php
```diff
@@ -144,131 +142,97 @@ public function appendTo(StreamName $streamName, Iterator $streamEvents): void
...
+        return $this->postLoad(__FUNCTION__ . '.post', $stream, $streamName, $event);
+    }
 
-        $this->getActionEventEmitter()->dispatch($event);
+    /**
+     * @throws Exception\StreamNotFoundException
+     */
+    public function loadReverse(
+        StreamName $streamName,
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

src/EventStore.php
```diff
@@ -144,131 +142,97 @@ public function appendTo(StreamName $streamName, Iterator $streamEvents): void
...
-    {
-        $argv = ['streamName' => $streamName, 'metadata' => $metadata, 'minVersion' => $minVersion];
-
-        $event = $this->actionEventEmitter->getNewActionEvent(__FUNCTION__ . '.pre', $this, $argv);
-
-        $this->getActionEventEmitter()->dispatch($event);
+    public function loadEventsByMetadataFrom(
+        StreamName $streamName,
+        array $metadata,
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

src/EventStore.php
```diff
@@ -144,131 +142,97 @@ public function appendTo(StreamName $streamName, Iterator $streamEvents): void
...
-     */
-    public function replay(array $streamNames, DateTimeInterface $since = null, array $metadatas = null): Iterator
-    {
-        if (empty($streamNames)) {
-            throw new Exception\InvalidArgumentException('No stream names given');
-        }
+    public function loadEventsReverseByMetadataFrom(
+        StreamName $streamName,
+        array $metadata,
+        int $fromNumber = 0,
```
codeliner: `int $fromNumber = 1`

tests/Aggregate/AggregateRepositoryTest.php
```diff
@@ -194,9 +194,14 @@ public function it_loads_the_entire_stream_if_one_stream_per_aggregate_is_enable
...
 
-        $adapter->load(Argument::that(function (StreamName $streamName) {
-            return $streamName->toString() === User::class . '-123';
-        }), null)->willReturn(new Stream(new StreamName(User::class . '-123'), new \ArrayIterator([])));
+        $adapter
+            ->load(
+                Argument::that(function (StreamName $streamName) {
+                    return $streamName->toString() === User::class . '-123';
+                }),
+                0,
```
codeliner: 1

tests/EventStoreTest.php
```diff
@@ -129,7 +127,7 @@ public function it_appends_events_to_stream_and_records_them(): void
 
         $secondStreamEvent = UsernameChanged::with(
             ['new_name' => 'John Doe'],
-            2
+            1
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -166,7 +164,7 @@ public function it_does_not_append_events_when_listener_stops_propagation(): voi
...
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -211,7 +209,7 @@ public function it_loads_events_by_matching_metadata(): void
 
         $streamEventWithMetadata = TestDomainEvent::with(
             ['name' => 'Alex', 'email' => 'contact@prooph.de'],
-            2
+            1
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -244,14 +242,14 @@ public function it_loads_events_by_min_version(): void
 
         $streamEventVersion2 = UsernameChanged::with(
             ['new_name' => 'John Doe'],
-            2
+            1
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -244,14 +242,14 @@ public function it_loads_events_by_min_version(): void
...
         );
 
         $streamEventVersion2 = $streamEventVersion2->withAddedMetadata('snapshot', true);
 
         $streamEventVersion3 = UsernameChanged::with(
-            ['new_name' => 'John Doe'],
-            3
+            ['new_name' => 'Jane Doe'],
+            2
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -274,7 +272,7 @@ public function it_loads_events_by_min_version(): void
         $this->assertTrue($loadedEventStream->streamEvents()[0]->metadata()['snapshot']);
         $this->assertFalse($loadedEventStream->streamEvents()[1]->metadata()['snapshot']);
 
-        $loadedEvents = $this->eventStore->loadEventsByMetadataFrom($stream->streamName(), [], 2);
+        $loadedEvents = $this->eventStore->loadEventsByMetadataFrom($stream->streamName(), [], 1);
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -290,6 +288,144 @@ public function it_loads_events_by_min_version(): void
...
+
+        $this->eventStore->beginTransaction();
+
+        $this->eventStore->create($stream);
+
+        $this->eventStore->commit();
+
+        $streamEventVersion2 = UsernameChanged::with(
+            ['new_name' => 'John Doe'],
+            1
```
codeliner: 2

tests/EventStoreTest.php
```diff
@@ -290,6 +288,144 @@ public function it_loads_events_by_min_version(): void
...
+        );
+
+        $streamEventVersion2 = $streamEventVersion2->withAddedMetadata('snapshot', true);
+
+        $streamEventVersion3 = UsernameChanged::with(
+            ['new_name' => 'Jane Doe'],
+            2
```
codeliner: 3

tests/EventStoreTest.php
```diff
@@ -290,6 +288,144 @@ public function it_loads_events_by_min_version(): void
...
+        );
+
+        $streamEventVersion3 = $streamEventVersion3->withAddedMetadata('snapshot', false);
+
+        $streamEventVersion4 = UsernameChanged::with(
+            ['new_name' => 'Jane Dole'],
+            3
```
codeliner: 4

tests/EventStoreTest.php
```diff
@@ -290,6 +288,144 @@ public function it_loads_events_by_min_version(): void
...
+
+        $this->eventStore->beginTransaction();
+
+        $this->eventStore->create($stream);
+
+        $this->eventStore->commit();
+
+        $streamEventVersion2 = UsernameChanged::with(
+            ['new_name' => 'John Doe'],
+            1
```
codeliner: 2

tests/EventStoreTest.php
```diff
@@ -290,6 +288,144 @@ public function it_loads_events_by_min_version(): void
...
+        );
+
+        $streamEventVersion2 = $streamEventVersion2->withAddedMetadata('snapshot', true);
+
+        $streamEventVersion3 = UsernameChanged::with(
+            ['new_name' => 'Jane Doe'],
+            2
```
codeliner: 3

tests/EventStoreTest.php
```diff
@@ -290,6 +288,144 @@ public function it_loads_events_by_min_version(): void
...
+        );
+
+        $streamEventVersion3 = $streamEventVersion3->withAddedMetadata('snapshot', false);
+
+        $streamEventVersion4 = UsernameChanged::with(
+            ['new_name' => 'Jane Dole'],
+            3
```
codeliner: 4

tests/EventStoreTest.php
```diff
@@ -337,7 +473,7 @@ public function it_returns_listener_events_when_listener_stops_loading_events_an
 
         $streamEventWithMetadata = UsernameChanged::with(
             ['new_name' => 'John Doe'],
-            2
+            1
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -351,7 +487,7 @@ public function it_returns_listener_events_when_listener_stops_loading_events_an
         $this->eventStore->getActionEventEmitter()->attachListener('loadEventsByMetadataFrom.pre', function (ActionEvent $event) {
             $streamEventWithMetadataButOtherUuid = UsernameChanged::with(
                 ['new_name' => 'John Doe'],
-                1
+                0
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -654,7 +790,7 @@ public function it_wraps_up_code_in_transaction_properly(): void
 
         $secondStreamEvent = UsernameChanged::with(
             ['new_name' => 'John Doe'],
-            2
+            1
```
codeliner: revert

tests/EventStoreTest.php
```diff
@@ -682,7 +818,7 @@ public function it_commits_transaction_if_adapter_implements_can_handle_transact
         $adapter->beginTransaction()->shouldBeCalled();
         $adapter->create($stream)->shouldBeCalled();
         $adapter->commit()->shouldBeCalled();
-        $adapter->load(Argument::any(), null)->willReturn($stream);
+        $adapter->load(Argument::any(), 0, null)->willReturn($stream);
```
codeliner: 1

tests/EventStoreTest.php
```diff
@@ -707,198 +843,11 @@ public function it_commits_transaction_if_adapter_implements_can_handle_transact
...
-
-        $this->eventStore->replay([new StreamName('user'), new StreamName('post')], null, [[]]);
-    }
-
     private function getTestStream(): Stream
     {
         $streamEvent = UserCreated::with(
             ['name' => 'Alex', 'email' => 'contact@prooph.de'],
-            1
+            0
```
codeliner: revert

src/EventStore.php
prolic: It returns Stream|array. That's why there is no return type here. Also it's a private method, so nobody has to deal with it.

src/EventStore.php
prolic: It returns Iterator|array. That's why there is no return type here. Also it's a private method, so nobody has to deal with it.

prolic:
Starting with event position 0 or starting with event positon 1 - that is the big question here :-)

Greg's event-store always starts with 0. So if we want to provide an adapter for it one day, we _have_ to start with 0, otherwise we run into problems. Or would you like to add an "StreamWasCreated" event on all streams at position 0 to fulfil Greg's requirements?

On array indexes we also start counting on 0, so using 0 is more compatible and makes also no problem at all, to start there for any other place. So why force it to be 1?

codeliner:
> So why force it to be 1?

Because a domain event with a version 0 makes no sense! 
Other reasons (but the first one is really the most important one):
- We would need to rewrite all our existing streams (but this seems to be required anyway)
- we need to align the logic in the EventSourcedAggregateRoot which makes migration even more hard as you need to coordinate code and database migration. Not sure if you can achieve this without downtime which would be really bad.

We have two PoVs her:
technical event store PoV: an index normally starts at 0
domain PoV: a domain event with version 0 makes no sense

Should we have both? Meaning the event version is not used for this, but instead we use event position internally? But this would make things even more complex :(

prolic:
> domain PoV: a domain event with version 0 makes no sense

So it's Greg's fault that his EventStore implementation makes no sense with event-sourcing? :-)

Two other things:

1) With event-position instead of version, we have one-stream-per-aggregate now. We could however add the version to the event metadata - so the other strategies would also work again. Having version in metadata also leads to, you don't need to have event-position === aggregate version (although I still stink an aggregate root is allowed to start with version 0).

2) That also means, that for not-one-stream-per-aggregate, you need to update the AggregateRepository, in that way, that you add the version-metadata to your query.

codeliner:
> So it's Greg's fault that his EventStore implementation makes no sense with event-sourcing? :-)

Don't know. My name is not Greg. I have my own thoughts :P

prolic:
So next question - no matter if we start counting with 0 or 1:

- If we want all current existing stream strategies, we need to update the AggregateRepository to add version-metadata to its queries.
- or: we use one-stream-per-aggregate - no other strategies (however you can have those things available to query with stream-to-stream-projections)

I'm for the latter - having one stream per aggregate! It also makes it much easier to understand for beginners. Remember all those questions about the strategies?

codeliner:
> I'm for the latter - having one stream per aggregate! It also makes it much easier to understand for beginners. Remember all those questions about the strategies?

Definitely agree here, but I don't think that one stream per aggregate work well for RDBMS, so we would need to drop the doctrine adapter. In traditional environments where DBAs control the RDBMS you cannot create tables on the fly! They won't grant you the rights for that. Also not sure if backup strategies work well with lets say 1.000.000 million tables ...
Supporting RDBMS is a really big selling point because you can introduce event sourcing step by step. Ask @rommson for example.
@basz as doctrine adapter maintainer what do you think?

basz:
Start with ordinality 1st, 2nd...

Symantically most correct

prolic:
@basz the question was to drop doctrine adapter :-)

codeliner:
@prolic 
> If we want all current existing stream strategies, we need to update the AggregateRepository to add version-metadata to its queries.

Does this option cause problems in other areas? We already have removed the strategies in favor of a simple `$oneStreamPerAggregate` flag in the repository. So we would just need to change the internals of the repository but keep support for all strategies. I'd prefer this option.

prolic:
So when we keep all current strategies and we add the version to the metadata - then we can also count the event position in the stream on 0, as the domain-event's version is not affected by this, right?

I don't see a technical problem with those strategies really, however I feel like remove it.

prolic:
One thing I just noticed: We have a Prooph\EventStore\Aggregate namespace in place. Does that make sense? Perhaps move this one to the event-sourcing repo?

codeliner:
> So when we keep all current strategies and we add the version to the metadata - then we can also count the event position in the stream on 0, as the domain-event's version is not affected by this, right?

sounds good to me

> One thing I just noticed: We have a Prooph\EventStore\Aggregate namespace in place. Does that make sense? Perhaps move this one to the event-sourcing repo?

agree here, too. And it is like removing the strategies because prooph/event-sourcing is just a suggestion ;)

basz:
Oh sorry - as I am coming from that beginner perspective i would not like to see the adapter go. It means one would need some other peersitence layer that must be maintained . Including the tools to inspect those.

So - for attracting new comers it would be sad if it can't work.

Then again - MySQL is not the fastest work so serious dev requires something else anyway.

I am not sure I understand why (Strategies) it would need to be dropped from a technical point of view. Naively saying it currently works.

And of course I can agree cause it would be the shortest main tenet career evar.. :)

codeliner:
hehe, this brings us back to the question if we should remove `version` from message interface? 
We can do that now because version is no longer required by the event-store but is only event metadata.
/cc @oqq

codeliner:
@basz 
> And of course I can agree cause it would be the shortest main tenet career evar.. :)

You're still maintainer of psr7-middleware ;)

> So - for attracting new comers it would be sad if it can't work.

That's the important fact ^^ and the compromise discussed with @prolic seems to be a good solution to support both use-cases.

prolic:
Ok to wrap up:

- This PR gets merged as is, unless someone finds a bug here. This means the event store starts at position 0 for all streams.
- The aggregate-namespace will be moved to the event-sourcing repository (another PR)
- $message->version() method will be removed (another PR)
- we keep doctrine-adapter (good for @basz)
- event version will be added to metadata
- we keep the stream strategies as part of the event-sourcing library

I try to submit the next 2 PR's with the mentioned changes tomorrow.

prolic:
Oh I nearly forgot: The InMemoryAdapter uses the $event->version() method to compare with event position. If we remove this here, I have to make another change so use the array-index for comparision - but that's subject to the next PR.

oqq:
> hehe, this brings us back to the question if we should remove version from message interface? 
We can do that now because version is no longer required by the event-store but is only event metadata.
/cc @oqq

haha :D
So, the version from a message is only relevant in context of her stream. Why not add the version in the stream instead? Which is simple the position from an iterator.

My 2 cents to the doctrine adapter. Here https://github.com/prooph/event-store/issues/196 i have mentioned i like to have a way to provide the stream name (and in this case also the table name for the doctrine adapter) dynamically per event stream / message. So if there is only one database table available for event stream, this service could return "event_stream" as table name. If i which to group my events per aggregate_type i can do so. Maybe this service could also be used to create database tables if necessary (or similar ways around to separate concerns). We could provide some implementations for that service. Maybe every store adapter provide her own naming strategies.
@codeliner

prolic:
Problem with naming strategies is this:
- Every database vendor has its own limitation in terms of forbidden characters - this makes migrating between databases hard.
- Every database vendor has its own limitation in table name string length / mongo collection string length, and so on - same problem of migration again

That's why I said, let's sha1 all event stream names and simply use a lookup table for the mapping between internal stream name and readable stream name.

If you like to append-only, you don't need this lookup or any strategy, just sha1 and you're safe.
Only when you create a stream, this lookup entry needs to be written, too.

But let's discuss this in the RFC here: https://github.com/prooph/event-store/issues/196.s

prolic:
I removed the aggregate namespace already in this PR. I need it quickly, so I can update the event-sourcing repository.

@codeliner ready for merge.

prolic:
I merged myself, so I can start working on event-sourcing repo. If there are any bugs left, another PR is needed.

PR #201, opened 2016-10-26
remove snapshot namespace

prolic:
@codeliner


prolic:
This will be moved to event-sourcing as well.

PR #203, opened 2016-11-03
add stream metadata

prolic:


prolic:
@codeliner merged as needed immediately

src/Adapter/InMemoryAdapter.php
```diff
@@ -34,11 +33,21 @@ class InMemoryAdapter implements Adapter
      */
     protected $streams;
 
+    public function fetchStreamMetadata(StreamName $streamName): ?array
+    {
+        if (isset($this->streams[$streamName->toString()]['metadata'])) {
+            return $this->streams[$streamName->toString()]['metadata'];
+        }
+
+        return null;
```
lucascourot: similar as `return $this->streams[$streamName->toString()]['metadata'] ?? null;`

PR #204, opened 2016-11-03
add metadata matcher

prolic:


src/EventStore.php
```diff
@@ -427,15 +428,15 @@ private function postLoad(string $action, ?Stream $stream, StreamName $streamNam
...
     ) {
         $argv = [
-            'streamName' => $streamName,
-            'metadata'   => $metadata,
-            'fromNumber' => $fromNumber,
-            'count'      => $count,
+            'streamName'        => $streamName,
+            'fromNumber'        => $fromNumber,
+            'count'             => $count,
+            'metadataMatcher'   => $metadataMatcher,
```
lucascourot: I wouldn't align values

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+        foreach ($this->data as $key => $value) {
+            if (! isset($metadata[$key])) {
+                return false;
+            }
+
+            $testValue = $this->data[$key]['value'];
+
+            switch ($this->data[$key]['operator']) {
+                case Operator::EQUALS():
+                    if ($testValue != $value) {
```
codeliner: strict type check please: !==

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+                        return false;
+                    }
+                    break;
+                case Operator::GREATER_THAN():
+                    if ($testValue <= $value) {
```
codeliner: Why not `$value > $testValue` ? This way it is confusing. You read GREATER_THAN but <= is used

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+                        return false;
+                    }
+                    break;
+                case Operator::GREATER_THAN_EQUALS():
+                    if ($testValue < $value) {
```
codeliner: `$value >= $testValue`

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+                        return false;
+                    };
+                    break;
+                case Operator::LOWER_THAN():
+                    if ($testValue >= $value) {
+                        return false;
```
codeliner: `$value < $testValue`

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+                    }
+                    break;
+                case Operator::LOWER_THAN_EQUALS():
+                    if ($testValue > $value) {
```
codeliner: `$value <= $testValue`

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Metadata;
+
+use Prooph\EventStore\Exception\InvalidArgumentException;
+
+class MetadataMatcher
```
codeliner: MetadataMatcher should be an interface and this class should be named ValueMetadataMatcher instead.
In other scenarios you may want to perform more complex matches like if this value is equal to xy, another value should be abc or if metadata key exists and so on

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+{
+    private $data = [];
+
+    public function data(): array
+    {
+        return $this->data;
+    }
+
+    public function addMetadataMatch(string $key, Operator $operator, $value): void
```
codeliner: I'd prefer immutable objects for this kind of objects to avoid side effects. A new instance of the  class should be returned and the method should be named `withMetadataMatch`

src/Metadata/MetadataMatcher.php
prolic: it should return false when the opposite happends, otherwise continue

prolic:
The metadata matching logic has been moved to the adapters, see: 5632d42746bbfa4e8994d073330c7bab8db3db2f

The the metadata matcher doesn't really match anymore, it's just a container. Better name is required I think :)

src/Metadata/MetadataMatcher.php
prolic: Ah you are right, I can change the order of the testing values :) my bad.

prolic: Still a problem here, the test reacts only to return false, I don't test for true, so the opposite of what is questioned in the matcher.

codeliner: I see. Could be solved in some way but yeah method is small enough to be understandable the way it is.

prolic: see: https://github.com/prooph/event-store/pull/205

src/Adapter/InMemoryAdapter.php
```diff
@@ -135,18 +136,22 @@ public function loadReverse(
...
             return new ArrayIterator();
         }
 
+        if (null === $metadataMatcher) {
+            $metadataMatcher = new MetadataMatcher();
+        }
+
         $streamEvents = [];
 
         foreach ($this->streams[$streamName->toString()]['events'] as $index => $streamEvent) {
```
oqq: Instead of using a foreach here we should use a filter method. The whole code would be more readable, since this spaghetti condition is just not readable. Also a method 'getEventsFromStream' in a StreamCollection would be nice.

src/Adapter/InMemoryAdapter.php
```diff
@@ -163,18 +168,22 @@ public function loadEvents(
...
```
oqq: Filter method vs. foreach with spaghetti condition

src/Metadata/MetadataMatcher.php
oqq: +1 for interface (+ final)

src/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,94 @@
...
+                default:
+                    throw new \UnexpectedValueException('Unknown operator found');
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * @param mixed $value
```
oqq: Maybe this could be omitted, since "no prefer" is always "mixed" ;)

src/Adapter/InMemoryAdapter.php
```diff
@@ -135,18 +136,22 @@ public function loadReverse(
 
     public function loadEvents(
         StreamName $streamName,
-        array $metadata = [],
         int $fromNumber = 0,
-        int $count = null
+        int $count = null,
+        MetadataMatcher $metadataMatcher = null
     ): Iterator {
         if (! isset($this->streams[$streamName->toString()])) {
```
oqq: I would prefer a hasStream method for readability. Maybe with a StreamCollection.

src/Adapter/InMemoryAdapter.php
```diff
@@ -163,18 +168,22 @@ public function loadEvents(
 
     public function loadEventsReverse(
         StreamName $streamName,
-        array $metadata = [],
         int $fromNumber = 0,
-        int $count = null
+        int $count = null,
+        MetadataMatcher $metadataMatcher = null
     ): Iterator {
         if (! isset($this->streams[$streamName->toString()])) {
```
oqq: I would prefer a hasStream method for readability. Maybe with a StreamCollection.

src/Metadata/MetadataMatcher.php
prolic: @oqq look at the current implementation in develop branch - this is only a container, there is no reason to have another implementation.

src/Adapter/InMemoryAdapter.php
prolic: wanna submit a PR?

src/Adapter/InMemoryAdapter.php
prolic: see: https://github.com/prooph/event-store/pull/207

src/Adapter/InMemoryAdapter.php
prolic: I don't know. Just submit a PR and make a suggestion, we'll see when we have it.

PR #208, opened 2016-11-04
removes nesting from recorded events

### LINKED ISSUE #202: Remove nesting from $recordedEvents
The current implementation for transactions stores all events from an aggregate in an $recordedEvents object. On every call to ```EventStore::create``` or ```EventStore::appendTo``` this object will be nested in itself.

@mikemix has provided an example where this could trigger xdebugs max nesting function level exception.
https://gist.github.com/mikemix/5ebb1726c42aeea633eca1feec1120e3

Maybe there is a better way to store this events, without an infinite nesting, which could have also positiv impact on performance.

I would suggest $recordedEvents is always an AppendIterator and got new events appended, instead of be nested on every round.

/cc @prolic

oqq:
This PR closes #202

### DIFF
--- a/src/EventStore.php
+++ b/src/EventStore.php
@@ -35,7 +35,6 @@ use Prooph\EventStore\Stream\StreamName;
 class EventStore
 {
     /**
-     *
      * @var Adapter
      */
     protected $adapter;
@@ -59,7 +58,7 @@ class EventStore
     {
         $this->adapter = $adapter;
         $this->actionEventEmitter = $actionEventEmitter;
-        $this->recordedEvents = new ArrayIterator();
+        $this->recordedEvents = new AppendIterator();
     }
 
     public function getAdapter(): Adapter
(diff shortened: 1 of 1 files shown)


prolic:
Can you add a test-case that is failing without this?

codeliner:
@prolic I guess a test-case is difficult here because the test would need to reach xdebugs max nesting function level and therefor xdebug would need to be installed and set up with an appropriate limit.

src/EventStore.php
```diff
@@ -359,7 +350,7 @@ public function rollback(): void
 
         $this->actionEventEmitter->dispatch($event);
 
-        $this->recordedEvents = new ArrayIterator();
+        $this->resetRecordedEvents();
```
sandrokeil: I'm not sure if it is worth it, to create a new method for a one line of code? I would prefer `$this->recorderEvents = new AppendIterator()`. Every function call in PHP is a performance drawback too.

prolic: :+1:

codeliner: @oqq In a domain model I would do it exactly like you: clean and readable code that tells you what it does. But @sandrokeil 's reasoning wins here, because prooph is infrastructure and performance has very high priority.

oqq: Of course, my reason was to be consistent what "recordedEvents" should be. But i agree with the performance reasons.

oqq:
It would be possible to implement a hackish test case, which would be skipped if xdebug is not available. But for me it is currently not possible to run xdebug with php7.1

prolic:
also manually merged into master

prolic:
see: https://github.com/prooph/event-store/releases/tag/v6.5.1

PR #217, opened 2016-11-19
use distinct event handler context in projections

prolic:


src/Projection/AbstractProjection.php
```diff
@@ -154,4 +150,60 @@ protected function handleStreamWithHandlers(string $streamName, Iterator $events
...
+                    $this->projection->linkTo($streamName, $event);
+                }
+
+                public function emit(Message $event): void
+                {
+                    $this->projection->emit($event);
+                }
+            };
+        } else {
+            return new class($this) {
```
prolic: thx, updated.

PR #219, opened 2016-11-21
Add stream name in handler context

prolic:


src/Projection/AbstractProjection.php
```diff
@@ -133,10 +131,13 @@ protected function handleStreamWithHandlers(string $streamName, Iterator $events
         foreach ($events as $event) {
             /* @var Message $event */
             $this->position->inc($streamName);
+
             if (! isset($this->handlers[$event->messageName()])) {
                 continue;
             }
+
             $handler = $this->handlers[$event->messageName()];
+            $handler = Closure::bind($handler, $this->createHandlerContext($streamName));
```
prolic: Or is this no performance problem at all?

prolic: Perhaps we add stream name as third argument to the handler, but this is something I like to avoid.

codeliner: handler context could be created outside of the `foreach` and with a `setStreamName` method but I would only do this if it has really positive influence on the performance.

codeliner: ah forget my idea. Immutability is king.

Issue #234, opened 2016-12-26
[RFC] Projection implementation


prolic:
Current state is that we have three interfaces (Query, Projection, ReadModelProjection) and the corresponding implementations. Besides that there are some abstractions for those being located in abstract classes and traits.

The problem is, that this might give too many things, users can extend from. It makes it more complicated and might lead to misuse and opened issues because of misunderstandings.

This RFC proposes several solutions to this. Let's discuss the options and find the best fit.

1) Remove all abstractions and have code duplication instead. Then we only have three final classes, where noone can extend from and do unintended things with.

2) Leave some abstractions in traits (they are uselss as is, as you cannot extend them).

3) Make them all anonymous classes inside the event store implementation. You can then only make use of if by calling `$eventStore->createQuery()` and similar calls. No direct instantiation is possible. This makes it completely impossible to extend in a weird way and make false usage. But this means also no abstractions at all are possible, leading to code duplication.

4) like in 3 but we add some traits to reduce code duplication.

5) Leave everything as is.

6) Your suggestions I didn't think of yet.

codeliner:
I'd rephrase 1.:

Remove all abstractions **but keep the interfaces** and have code duplication instead. Then we only have three final classes **(suggested implementation by prooph team) + interfaces if someone want to do it differently**. So noone can extend from and do unintended things with.

sandrokeil:
Agree with @codeliner (we should provide interfaces) but I would also use Traits if it makes sense and userland code can use them.

sandrokeil:
@prolic Do you know or can you imagine if a graph db has any benefits for projections? Maybe we can consider this in the interfaces?

prolic:
Of course we keep the interfaces - it was not up for vote to remove them ;-) Sorry if I was unclear on this. 

About traits: Userland code should not use them at all. A trait would be only used internally to reduce code duplication. We could mark them with `@internal` in phpdoc and comment with don't use these traits on your own, unless you are implementing the event-store interface yourself.

About graph dbs: There are a lot of useful scenarios for them. Currently I don't see what should be added for this. The only relevant methods to create a read model with the ReadModelProjection is by calling

```
public function stack(string $operation, ...$args): void;

// and

public function persist(): void;
```

You can stack whatever operation you want, so the possibilities are endless and there is nothing needed to add, unless you provide an example. See also: https://github.com/prooph/event-store/blob/develop/src/Projection/ReadModel.php

codeliner:
solved by #239

Issue #236, opened 2017-01-01
RFC projections: if no new events found at all, sleep for 0.25 secs


codeliner:
As discussed in the chat we should add a configurable sleep time to projections. Atm the `keepRunning` flag causes a lot of load even if no new events are added to a stream. With a sleep time we can avoid this issue.

sandrokeil:
We should use microseconds and `usleep`.

codeliner:
solved by #239

PR #239, opened 2017-01-03
Remove abstract classes in projections + sleep when no new events found + projection factories

prolic:
resolves:
https://github.com/prooph/event-store/issues/234
https://github.com/prooph/event-store/issues/236
https://github.com/prooph/event-store/issues/240

- remove position class
- remove all abstract classes and traits for projections
- if no new events found at all, sleep for 0.25 secs
- read model projections can no longer `emit` and `linkTo` - because this makes no sense when creating a read model
- add projection factories


prolic:
ping @codeliner

prolic:
@oqq review?

codeliner:
LGTM

@oqq second review would be good

prolic:
@codeliner updated for https://github.com/prooph/event-store/issues/240

prolic:
the line https://github.com/prooph/event-store/blob/remove_abstract_classes_in_projections/src/Projection/InMemoryEventStoreProjection.php#L105 has no code coverage, although the inner block is tested, I don't know, a problem with xdebug maybe?

src/ActionEventEmitterEventStore.php
```diff
@@ -304,22 +307,62 @@ public function updateStreamMetadata(StreamName $streamName, array $newMetadata)
...
-        return $this->eventStore->createProjection($name, $options);
+    public function createProjection(
+        string $name,
+        ProjectionOptions $options = null,
+        ProjectionFactory $factory = null
+    ): Projection {
+        if (null === $factory) {
+            $factory = $this->getDefaultProjectionFactory();
+        }
+
```
codeliner: shouldn't the ActionEventEmitterEventStore also invoke `$factory->setEventStore($this);`, not only when creating the default factory?

src/ActionEventEmitterEventStore.php
```diff
@@ -304,22 +307,62 @@ public function updateStreamMetadata(StreamName $streamName, array $newMetadata)
...
         ReadModel $readModel,
-        ProjectionOptions $options = null
+        ProjectionOptions $options = null,
+        ReadModelProjectionFactory $factory = null
     ): ReadModelProjection {
-        return $this->eventStore->createReadModelProjection($name, $readModel, $options);
+        if (null === $factory) {
+            $factory = $this->getDefaultReadModelProjectionFactory();
+        }
+
```
codeliner: same question as above

src/ActionEventEmitterEventStore.php
```diff
@@ -304,22 +307,62 @@ public function updateStreamMetadata(StreamName $streamName, array $newMetadata)
...
+        return $factory->factory($name, $readModel, $options);
+    }
+
+    public function getDefaultQueryFactory(): QueryFactory
+    {
+        $factory = $this->eventStore->getDefaultQueryFactory();
+        $factory->setEventStore($this);
```
codeliner: move to `createQuery` ?

src/ActionEventEmitterEventStore.php
```diff
@@ -304,22 +307,62 @@ public function updateStreamMetadata(StreamName $streamName, array $newMetadata)
...
+
+        return $factory;
+    }
+
+    public function getDefaultProjectionFactory(): ProjectionFactory
+    {
+        $factory = $this->eventStore->getDefaultProjectionFactory();
+        $factory->setEventStore($this);
```
codeliner: move to `createProjection`

src/ActionEventEmitterEventStore.php
```diff
@@ -304,22 +307,62 @@ public function updateStreamMetadata(StreamName $streamName, array $newMetadata)
...
+
+        return $factory;
+    }
+
+    public function getDefaultReadModelProjectionFactory(): ReadModelProjectionFactory
+    {
+        $factory = $this->eventStore->getDefaultReadModelProjectionFactory();
+        $factory->setEventStore($this);
```
codeliner: move to createReadModelProjection

prolic:
@codeliner updated!

src/Projection/InMemoryEventStoreQuery.php
```diff
@@ -12,16 +12,345 @@
...
+    /**
+     * @var ?string
+     */
+    private $currentStreamName = null;
+
+    public function __construct(EventStore $eventStore)
+    {
+        $this->eventStore = $eventStore;
+
+        if ($eventStore instanceof ActionEventEmitterEventStore) {
```
codeliner: I'm with @oqq (see https://github.com/prooph/event-store/issues/240#issuecomment-270326292), that a `EventStoreDecorator` interface would be useful for this scenario here.
Maybe a user wants to use a custom event store decorator in this case the factory would not detect it.

Thoughts? we can add this interface in a separate PR. This one is already large enough :D

src/Projection/InMemoryEventStoreQueryFactory.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * @var EventStore
+     */
+    private $eventStore;
+
+    public function __construct(EventStore $eventStore)
+    {
+        $this->eventStore = $eventStore;
+    }
+
```
sandrokeil: @prolic Why not using a callable only? Is the setter really needed? /cc @codeliner 

```
public function __invoke(ProjectionOptions $options = null) {

}
```

prolic: without the setter, where should `$this->eventStore` come from?

sandrokeil: @prolic The event store is mandatory in constructor. Do I miss something?

```php
    public function __construct(EventStore $eventStore)
    {
        $this->eventStore = $eventStore;
    }
```

A better solution is this, right? No constructor, no setter, no cry. :-) This would be a clean interface, same for the other factories.

```php
interface QueryFactory
{
    public function __invoke(EventStore $eventStore, ProjectionOptions $options = null): Query;
}
```

prolic: okay, wanna provide a PR?

src/InMemoryEventStore.php
```diff
@@ -283,39 +286,53 @@ public function transactional(callable $callable)
...
-            $options->sleep()
-        );
+        return $factory->factory($name, $readModel, $options);
+    }
+
+    public function getDefaultQueryFactory(): QueryFactory
+    {
+        return new InMemoryEventStoreQueryFactory($this);
+    }
+
```
sandrokeil: @prolic We don't need everytime a new instance (especially with invoke), should be cached. What do you think?

prolic: ok, put in same PR if you want to provide one

src/Projection/InMemoryEventStoreQueryFactory.php
sandrokeil: Will do later today.

prolic: thx :+1:

PR #238, opened 2017-01-02
outsourced configuration from ConfigurableAggregateTranslator

dropdevcoding:
some more tests and docs have to be added

PR #241, opened 2017-01-03
Revert "outsourced configuration from ConfigurableAggregateTranslator"

codeliner:
Reverts prooph/event-store#238


codeliner:
@prolic Can you take a second look? travis is red due to docheader checks but tests are green. would like to revert first and then correct the headers.

prolic:
all good. it's 2017 now, that's why travis is complaining

PR #246, opened 2017-01-06
Improve projection factories, use a simple callable definition

sandrokeil:
I've updated the Exception tests, because with this implementation, no wrong EventStore can be injected. Since the factories are callables, the event store is provided to the factory by the function itself.

Added lazy intializing of default factories. This has a better performance, instead of creating the factory by each call. The Factory is also stateless now.


prolic:
thx :+1:

PR #247, opened 2017-01-07
update event store decorator + query factory

prolic:


prolic:
ping @codeliner

prolic:
merged because I need it now to finish pdo-event-store.

if something is to discuss, let's open another PR afterwards

Issue #252, opened 2017-01-27
RFC: EventStore::load method changes


prolic:
The current implementation of EventStore v7 load / loadReverse method returns an instance of `Stream`.

This delivers 3 informations:
a) The stream name
b) The requested events
c) The stream metadata

For performance reasons the metadata is not queried in the result response, so if you want to get the metadata, you have to call the `fetchStreamMetadata`-method. This is the first problem with the load / loadReverse method.

Second, the stream name is already known, it's the first argument to the method.

Third, the only information that we get, that we don't have yet, are the events itself. Also calling `$eventStore->load(/* some params*/)->streamEvents()` everytime is a little too much, why is the load-method not returning the stream events directly?

I propose therefore, that the EventStore::load / loadReverse methods return the event iterator directly, not an instance of `Stream`.

As an alternative, we could also implement `IteratorAggregate` in the `Stream`, so at least you can iterate over the stream directly.

Please leave your vote below (A, B, C):

A - change load / loadReverse method to return in iterator directly
B - implement `IteratorAggregate` in the `Stream`
C - do nothing, leave it as is

/cc @basz @bweston92 @sandrokeil @oqq @codeliner @MattKetmo

For current interface see: https://github.com/prooph/event-store/blob/develop/src/ReadOnlyEventStore.php

oqq:
B

But also I would like to make the `Stream` object lazy, so that it could provide metadata and events. The underlying requests should be done on first `next()` call on the provided iterator or by calling `metadata()`. Maybe to much over engineered, but a `Stream` object should  known all stream relevant data.

codeliner:
I continue my "simplify all the things" trip so +1 for A

prolic:
Me also for A

basz:
sorry, no opinion

bweston92:
A seems reasonable. Mainly because like you've said you don't want to keep calling methods to get the same goal as expected when calling one method "load".

PR #253, opened 2017-01-29
Change EventStore::load / loadReverse methods

prolic:
resolves https://github.com/prooph/event-store/issues/252


prolic:
If this is accepted, I'll try to update the pdo-event-store accordingly asap.

Issue #254, opened 2017-02-06
Add support for inline upcasting


mrook:
Upcasters are classes that take one input event of revision `x` and output zero or more new events of revision `x + 1`.

They should be executed inline when loading event streams from the event store. Upcasters can possibly be implemented as event store plugins.

Possible scenarios:

- single event to single event
- single event to multiple (or zero) events
- chaining multiple upcasters

Inspiration: http://www.axonframework.org/docs/2.0/repositories-and-event-stores.html#event-upcasting (and https://docs.axonframework.org/v/3.0/part3/repositories-and-event-stores.html for the new version).

codeliner:
Sounds interesting, thx

prolic:
@mrook see https://github.com/prooph/event-store/pull/255

PR #255, opened 2017-02-07
Upcasting

prolic:
resolves https://github.com/prooph/event-store/issues/254

### DIFF
--- a/src/Upcasting/UpcasterChain.php
+++ b/src/Upcasting/UpcasterChain.php
@@ -1,0 +1,37 @@
+<?php
+/**
+ * This file is part of the prooph/event-store.
+ * (c) 2014-2017 prooph software GmbH <contact@prooph.de>
+ * (c) 2015-2017 Sascha-Oliver Prolic <saschaprolic@googlemail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Upcasting;
+
+use Prooph\Common\Messaging\Message;
+
+final class UpcasterChain implements Upcaster
+{
+    /**
+     * @var Upcaster[]
+     */
+    private $upcasters;
+
+    public function __construct(Upcaster ...$upcasters)
+    {
+        $this->upcasters = $upcasters;
+    }
+
+    public function upcast(Message $message): Message
+    {
+        foreach ($this->upcasters as $upcaster) {
+            $message = $upcaster->upcast($message);
+        }
+
+        return $message;
+    }
+}
--- a/src/Upcasting/UpcasterChain.php
+++ b/src/Upcasting/UpcasterChain.php
@@ -26,12 +26,21 @@ final class UpcasterChain implements Upcaster
         $this->upcasters = $upcasters;
     }
 
-    public function upcast(Message $message): Message
+    public function upcast(Message $message): array
     {
+        $result = [];
+        $messages = [$message];
+
         foreach ($this->upcasters as $upcaster) {
-            $message = $upcaster->upcast($message);
+            $result = [];
+
+            foreach ($messages as $message) {
+                $result += $upcaster->upcast($message);
+            }
+
+            $messages = $result;
         }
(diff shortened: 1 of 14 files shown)


prolic:
For discussion:

- We need an instance of a message to create a new message - this means when updating events, you need to create a new class, f.e. `UserRegistered` => `UserRegisterd` + `UserRegisteredV2` - is that okay?
- Should we maybe prefer to do upcasting with message name / payload / metadata only? If should remove the message factory from the event store, add our own (maybe `GenericMessageFactory`) - and use the `real` MessageFactory only within this plugin - how can this be achieved, if this is preferred?
- Do we need support for looking into the future of the stream per upcasting? Or can the upcaster itself store some information internally is required?

mrook:
That was fast! I'll take an in-depth look as soon as I have time.

As for the discussion points:

- the new event class needs to exist before the upcaster can work.
- the question is whether you want each upcaster individually to handle deserializing the raw event (so that would indeed mean only passing metadata & payload), or present the deserialized domain event to the upcaster; I think I prefer the first version, as serialization may have changed over time as well.
- some upcasters could need this to merge events, they may be more interested in the past than the future however. In general I think this is something best handled in individual upcasters.

prolic:
@codeliner @mrook tests now included

jkuchar:
Why not put upcast and downcast methods directly into events? These up/downcast methods are tightly coupled with event in particuar version anyway...

prolic:
@jkuchar Single Responsibility Principle

mrook:
@jkuchar because some upcasters would need information from other events (i.e., if you are merging) or emit multiple events (if you are splitting).

src/Upcasting/UpcasterChain.php
```diff
@@ -0,0 +1,46 @@
...
+     * @var Upcaster[]
+     */
+    private $upcasters;
+
+    public function __construct(Upcaster ...$upcasters)
+    {
+        $this->upcasters = $upcasters;
+    }
+
+    public function upcast(Message $message): array
```
codeliner: Can you take a second look on the message?

Why is the message turned into a messages array just to foreach over it?

You wanted to return `messages`, right? But `$result` is returned.
Foreach can be removed in either case.

prolic: You're right. I will fix this later.

src/Upcasting/UpcastingIterator.php
```diff
@@ -0,0 +1,91 @@
...
+    public function current(): ?Message
+    {
+        if (! empty($this->storedMessages)) {
+            return reset($this->storedMessages);
+        }
+
+        $messages = $this->upcaster->upcast($this->innerIterator->current());
+
+        if (empty($messages)) {
+            $this->next();
```
prolic: $this->innerIterator->next();

src/Upcasting/UpcastingIterator.php
```diff
@@ -0,0 +1,91 @@
...
+
+            if ($this->valid()) {
```
prolic: $this->innerIterator->valid()

src/Upcasting/UpcasterChain.php
prolic: Sorry, you are `not` right :)

At the first iteration of upcasters, you would have been right, but in the second iteration, I might have an array of messages, that may need further upcasting. see: UpcasterChainTest::it_chains_upcasts

This test would immediately break, if we would make the suggested change.

prolic:
There is still a bug in the iterator... I am trying to fix this now.

prolic:
@codeliner I fixed the upcasting iterator

prolic:
@mrook @codeliner your thoughts in the implementation?

mrook:
@prolic looks fine for the initial implementation!

src/Upcasting/UpcasterChain.php
codeliner: ok, I see. Just a few lines of code but very tricky ;)

codeliner:
LGTM

mrook:
:+1:

PR #256, opened 2017-02-09
Replace method_exists() calls with is_callable() to fix #235

### LINKED ISSUE #235: ConfigurableAggregateTranslator: method_exists vs. is_callable
While experimenting with the event store I ran into the following issue - I actually wanted to propose a PR but since it involves a bit of refactoring of the unit tests I thought it might be better to discuss the issue first before pouring in a lot of work ;)

In my test setup I used the following code from the documentation to set up a repository:
```php
$this->repository = new AggregateRepository(
    $this->eventStore,
    AggregateType::fromAggregateRootClass(\My\Domain\Model\User:class),
    new ConfigurableAggregateTranslator()
);
```

Which led to the following exception:
```
Call to protected method Prooph\EventSourcing\AggregateRoot::popRecordedEvents() from context 'Prooph\EventStore\Aggregate\ConfigurableAggregateTranslator'
```

The exception is correct: The ConfigurableAggregateTranslator tried to access the popRecordedEvents() which I did not made public, it still had the `protected` visibility as defined in \Prooph\EventSourcing\AggregateRoot.

The exception I would have expected in this case is:
```
Can not pop recorded events from aggregate root My\Domain\Model\User. The AR is missing a method with name popRecordedEvents!
```

The exception did not get fired because of the usage of `method_exists`. The function will check - as the name implies - if a method exists but not if it actually can be called from the current context. Thus it might be better to rely on a `is_callable` check. `is_callable` will take the current context into account.

Whilst changing the code of ConfigurableAggregateTranslator by replacing every `method_exists` call with `is_callable` is simple, a few unit tests will fail and need to be fixed. The problem is that the Prophecy magic relies on the `__call()` magic method which will make `is_callable` return true for methods that do not exist. Thus instead of using mocked objects the unit tests would need to work with simple dummy objects.

Happy to provide a PR to fix the issue but as I pointed out above I wanted to get your feedback first if this problem is in fact a problem or if it is ok the way it is ;)

shochdoerfer:


prolic:
Thanks :+1:

PR #259, opened 2017-02-13
Add testCase to handle stream only with new events

rabbl:
I added a testcase to increase the linecoverage in the eventStoreProjection.
If there are no new events, a NoNewerEventsFound-Exeption will be thrown and catched.
So the loop continues checking the next stream.
Please note, there is a possible case, that the projection cannot be stopped.

### DIFF
--- a/src/InMemoryEventStore.php
+++ b/src/InMemoryEventStore.php
@@ -15,6 +15,7 @@ namespace Prooph\EventStore;
 use ArrayIterator;
 use Iterator;
 use Prooph\Common\Messaging\Message;
+use Prooph\EventStore\Exception\NoNewerEventsFound;
 use Prooph\EventStore\Exception\StreamExistsAlready;
 use Prooph\EventStore\Exception\StreamNotFound;
 use Prooph\EventStore\Exception\TransactionAlreadyStarted;
@@ -149,6 +150,10 @@ final class InMemoryEventStore implements TransactionalEventStore
             ) {
                 $streamEvents[] = $streamEvent;
             }
+        }
+
+        if (count($streamEvents) == 0) {
+            throw NoNewerEventsFound::with($streamName);
         }
 
         return new ArrayIterator($streamEvents);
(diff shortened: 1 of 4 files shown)


src/Projection/InMemoryEventStoreProjection.php
```diff
@@ -328,8 +328,7 @@ public function run(bool $keepRunning = true): void
             foreach ($this->streamPositions as $streamName => $position) {
                 try {
                     $streamEvents = $this->eventStore->load(new StreamName($streamName), $position + 1);
-                } catch (Exception\StreamNotFound $e) {
-                    // no newer events found
+                } catch (Exception\NoNewerEventsFound $e) {
```
rabbl: The isStoped-conditional could go outside of the foreach-loop.

src/InMemoryEventStore.php
```diff
@@ -151,6 +152,10 @@ public function load(
             }
         }
 
+        if (count($streamEvents) == 0) {
+            throw NoNewerEventsFound::with($streamName);
```
prolic: why is this needed? i am okay with returning a empty iterator.

rabbl:
The InMemoryEventStoreTest is failing now, because the the throws exception is unexpected.
I can fix this, but I'm not very convinced that I did it in the right way.

src/InMemoryEventStore.php
rabbl: I have tried to write a test to increase the linecoverage here:
https://coveralls.io/builds/10114395/source?filename=src%2FProjection%2FInMemoryEventStoreReadModelProjection.php#L298
As I could see, there was no test possible to throw this exception.
So my idea war to add an eventsNotFound exception instead and catch it.

rabbl: Whats your idea in this case?

prolic: @rabbl this exception catching can be removed completely. No StreamNotFound exception is thrown any more, when no newer events are found - an empty iterator is returned instead, so the catching can be removed. wanna PR this?

rabbl: Done! 
https://github.com/prooph/event-store/pull/260

PR #260, opened 2017-02-13
Update test it_resumes_projection_from_position

rabbl:
In this test the annotation is missing.
The projection has to run with the keepRunning-param = false.


prolic:
Thanks :+1:

PR #261, opened 2017-02-13
allow to fetch stream names, category names and projection names

prolic:
This allows to get a filtered, paginated list of stream names, category names and projection names - especially useful for event-store-ui.

Regex is even allowed!

/cc @codeliner @oqq @basz


codeliner:
nice

Issue #265, opened 2017-02-18
Make Projections more abstract


oqq:
The current base implementation of ProjectionOptions has some mandatory values with defaults, which would not be used in every projection class. There is also an inheritance in PdoEventStoreProjection to add more options. To provide one value other than the defaults, I have to investigate the default values in the class and add they by myself in the constructor, otherwise the object could not be build.

I guess this would escalate.

My proposal is to make this DTO more abstract, so it can be used in every current and future implementation of projection and here factories.

Default values required by the factories should be defined by the factory and not in an abstract DTO.
Also, every projection should know which options are mandatory.

Here is an example implementation to be more clear what I am talking about:
https://gist.github.com/oqq/c709eae6d993b49d888d15703a466f30

This would be a bc break and should not implemented in an already beta version. But I guess there is not an implementation for projections out there other than from us.

What do you think about that?
@prolic @codeliner

prolic:
We did not release a final version, so BC breaks are still acceptable.

With the projection options that abstract, we can also use arrays for this purpose, there is no need to have a class anymore. But what I dislike, is the huge amount of constants added to the event-store implementation. Currently I am note sure, what's the best way to go.

oqq:
Sure an array would also be fine. 

But in that way we have a tool to provide some recurring instructions how: give me that or the default, test that array against my mandatories. Maybe this DTO could handle some of the constants which would be recurring over most projectors. Where the constants are is not important. But the default values should be in constants instead of method argument defaults. Also for investigations where this values would be used constants are much better than simple strings for array keys. I like it to jump in PhpStorm around by clicking and searching for implemenations instead of searching for a string in whole code.

sandrokeil:
Why does we not use an `interop-config` factory for the options. Or why not use an `interop-config` factory for the projections? An array as options is maybe not the best choice for mandatory constructor args.

I guess the best choice is to use a factory which implements `interop-config` for the projections so the config can also be [generated](https://sandrokeil.github.io/interop-config/reference/console-tools.html). And simple check with assertions in contructor for mandatory fields in the options array. The factory can also define the options as constants or we use a specific option class for this.

```
PdoEventStoreProjectionOptions::OPTION_PDO_CONNECTION

or

PdoEventStoreProjectionFactory::OPTION_PDO_CONNECTION


public function __construct(EventStore $eventStore, string $name, array $options)
```

oqq:
In that way the factory could be bypassed and we have to re-test all values in the projection, right?

[...]

PR #266, opened 2017-02-19
add methods to fetch projection state, status and stream positions

prolic:
Another nice addition for event-store-http-api / -ui / -monitoring


prolic:
Please approve only and do not merge. I like to prepare the pdo-event-store implementation, once this is accepted and then merge both at the same time, to avoid problems with people currently using the -dev version.

codeliner:
@prolic Changes are approved, but due to the ongoing discussion in #265 we should wait with further PRs until we have a result

Issue #257, opened 2017-02-11
Provide abstract event store test


oqq:
I am, as a redis event store developer, have to copy integration tests from pdo event store and hold fingers crossed that all tests would be enough.

Maybe someone could provide an abstract test case which could be used to test other implementations?

Maybe someone who really knows what should going on. ;)

prolic:
I looked into this and this is really hard to do. Same parts are really dependent on the pdo-event-store implementation details and cannot be reused for other implementations. Some other stuff is not possible with InMemoryEventStore (just because it's in memory and not persistent).

What I can suggest, you may create a common AbstractTestCase that can be inherited from, but keep in mind, the abstract test case cannot be a complete event store test-suite, as some things are implementation dependent to test (like querying the pdo-connection, dropping / recreating tables, etc).

prolic:
@oqq Do you want to provide this?

oqq:
Sure.. this TestCase was not to be a full test. Only some default should be reflected.
- if I create a stream with events, load should return this events.

Its more an integration test. Every EventStore still need his own unit tests.

Yes I can get into this. With some assistance and tips from you in PRs. ;)

prolic:
I take over.

prolic:
transactional event store test trait and abstract projection manager test will also be added

PR #267, opened 2017-02-20
Projection manager

prolic:
- [x] added projection manager
- [x] removed projection options class
- [x] projection options constants
- [x] added methods to fetch projection state, status and stream positions
- [x] added common event store test for all implementations
- [x] added common transactional event store test trait for all implementations
- [x] added common projection manager test for all implementations
- [x] added common event store projection test for all implementations
- [x] added common event store query test for all implementations
- [x] added common event store read model projection test for all implementations
- [x] updated all fetch*-methods
- [x] provided more tests
- [x] some bug fixes

supersedes https://github.com/prooph/event-store/pull/266

resolves https://github.com/prooph/event-store/issues/265

resolves https://github.com/prooph/event-store/issues/257

/cc @oqq


prolic:
oh and please only approve this PR, do not merge yet. I want to make the same changes to pdo-event-store before merging, so nothing breaks for current users. We can then merge both repositories at the same time.

src/Projection/InMemoryProjectionManager.php
```diff
@@ -0,0 +1,195 @@
...
+
+        $skipped = 0;
+        $found = 0;
+
+        $projectionNames = array_keys($this->projections);
+        ksort($projectionNames);
+
+        foreach ($projectionNames as $projectionName) {
+            if ($regex) {
+                if (! preg_match("/$filter/", $projectionName)) {
```
oqq: Maybe preg_grep is better instead of preg_match in a loop.
http://php.net/manual/en/function.preg-grep.php

src/Projection/InMemoryProjectionManager.php
```diff
@@ -0,0 +1,195 @@
...
+        throw new Exception\RuntimeException('Stopping a projection is not supported in ' . get_class($this));
+    }
+
+    public function fetchProjectionNames(?string $filter, bool $regex, int $limit, int $offset): array
+    {
+        if (null === $filter && $regex) {
+            throw new Exception\InvalidArgumentException('No regex pattern given');
+        }
+
+        if ($regex && false === @preg_match("/$filter/", '')) {
```
oqq: If you use a single call with preg_grep with an try/catch by using set_error_handler, you can avoid that @ and a second regex parse process.

src/Projection/ProjectionManager.php
```diff
@@ -0,0 +1,46 @@
...
+    public function deleteProjection(string $name, bool $deleteEmittedEvents): void;
+
+    public function resetProjection(string $name): void;
+
+    public function stopProjection(string $name): void;
+
+    /**
+     * @return string[]
+     */
+    public function fetchProjectionNames(?string $filter, bool $regex, int $limit, int $offset): array;
```
oqq: How about `fetchProjectionNames` and `fetchProjectionNamesRegex` instead of if/else conditions and this $regex flag?

prolic:
A factory for the InMemoryProjectionManager is still needed, but not yet implemented, will be part of another PR.

prolic:
Having a generic projection manager is not possible, because you always need the implementation details at hand, when doing some operations. I tried to already :)

src/Projection/ProjectionManager.php
prolic: I am okay with this, but this adds even more methods. @codeliner your thoughts?

src/Projection/InMemoryProjectionManager.php
prolic: PR against my branch?

src/Projection/InMemoryProjectionManager.php
prolic: PR against my branch?

prolic:
Copied from the chat (referring generic project manager):

> @oqq Have you an example which implementation details are needed?
> @prolic 1) MySqlProjectionManager has additional options (OPTION_LOCK_TIMEOUT_MS)
2) createReadModelProjection is much easier to call, with fewer parameters, than to create projection manually, the projection manager has to know how the projection is created
3) deleteProjection needs projection table
4) fetchProjectionNames and others needs projection table
> @oqq Currently you would add options to create projections in ProjectionManager. This Manager has to be build in a factory. addProjection is not needed, sure. But why not build the Projection instances in the factory and add they right in the manager factory?
So the manager is aware about all projections and could be generic.
AND the manager could handle different projection implementations at once.
If you worry about performance create projections with a lazy callable. :wink:
So we need one ProjectionManager in the event-store package. No need to implement such a manager for every projection or event store package.
> @prolic so what you want is:

$projection = $factory(.....);

instead of

$projection = $projectionManager->createProjection(.....);

if so, how do you call $projectionManager->deleteProjection('some_name', false); ?
> @oqq new ProjectionManager($eventStore, [ 'projection_1' => $projectionInstance or $projectionFactoryCallable ],
> @prolic that's what I just explained
it will not work!
$manager->deleteProjection() !== $projection->delete()
look at pdo-event-store::deleteProjection and pdo-projection::delete
they are not the same !
> @oqq I know what you would saying. I see no reason why the projection should not be aware how to delete himself. Maybe ::delete was the false method call here. Than it should be ::registerDeleteRequest
> @prolic also this means, that I need to have access to the projection code in order to stop / reset / delete it, which is not true. the projection can live in another microservice and have its own repository, so I cannot stop a projection, without having full access to its source code and configuration
in current implementation, I can stop / reset / delete any projection by name, without having the read model and other configuration at hand
> @oqq The Projection always knows where projections infos are stored, right?
The PdoProjection updates the projectionTable..
All methods on the manager deals only with projection names.. so it knows which projection instance belongs to that name.
registerDeleteRequestis not the best way. But setStatus should work
btw.. in my applications I would never store projection states beside the event store.
Currently the event store must never be touched. As DevOp I can secure this storage at the best and provide got backup strategies.
If a projection was corrupted or damaged, I can reset the whole read model infrastructure and dont have to touch the event store infrastructure at all.
So I would prefer to store the projection state beside the read model storage.
More than that.. I reckon my DevOp would request that.
> @prolic Imagine you had an even store ui, for the setup you need only the event store and projections table. There you want to stop a running projection. It is impossible to do that, if your projection state storage is on another database and you have to know how to build that projection upfront, in order to stop it.
> @oqq You always need to know how the projection is build, since you injects the projectionTable to the EventStore AND to the projection. If you use another MySQL Server or another projection table in event store ui, projection stopping would also not work.
But i understand what your issue is..
Is there an implementation of stopping projections in the event store ui?
I need a look on that.. maybe I have a proposal how to fix that issue.
Yay.. but we need again a ProjectManager for every EventStore implementation. And the Projection state must be stored with same method how EventStore stores values.
I am really against that, reasons above.
But honestly I am not sure how to implement a event store ui without knowing the whole projection infrastructure.
> @prolic You want the projection table to be in another database? This would mean injecting two pdo connections into the manager
> @oqq I would store the projection state in an own storage, yep.
Maybe nearby the read model stores.. maybe somewhere else.
Nope.. this means the manager has to known only projections.
If my read models stores values in mongo db.. than the projection state could also be stored in mongo
> @prolic With no benefit except making the manager more complex
> @oqq I have infrastructure benefits.. and the manager would not be complex.. he would less complex and more generic.
i remove the createProjection method
> @prolic It is not possible to have a generic projection manager, also with removed createProjection method
> @oqq The manager already knows about all projections (or is extendabl, doesn't matters)
why not? :smiley:
> @prolic You cannot simply add all projections available in different repos and microservices to the manager
There is a boundary 
> @oqq But what happens if I call createProjection?
> @prolic You create one and have obviously access to the needed deps in place
> @oqq So I can create projections I know, but can also stop projections I not know?
> @prolic yes
> @oqq Ok now I understand why you would couple eventstore and projections.
I really like the new way to remove the projection factories from projections. We should not pull away from that.

prolic:
InMemoryProjectionManagerFactory added

prolic:
Also from chat:

> @oqq ProjectionManager::__construct(EventStore, ProjectionStatusStore) :wink:
please don´t bash me :smiley:
Oh.. the projection manager does not need to know about event store.
ProjectionManager::__construct(ProjectionStatusStore)
> @prolic more interfaces, more implementations, is the projection status store another repostory? if so, more composer dependencies

prolic:
Btw: I already started the pdo-event-store upgrade, but there is much more to do, so this will take 2-3 more days.

src/Projection/ProjectionManager.php
codeliner: sounds good

codeliner:
LGTM

Need to read the chat again tomorrow. too late today ;)

src/Projection/InMemoryProjectionManager.php
```diff
@@ -0,0 +1,195 @@
...
+final class InMemoryProjectionManager implements ProjectionManager
+{
+    public const OPTION_CACHE_SIZE = 'cache_size';
+    public const OPTION_SLEEP = 'sleep';
+    public const OPTION_PERSIST_BLOCK_SIZE = 'persist_block_size';
+
+    private const DEFAULT_CACHE_SIZE = 1000;
+    private const DEFAULT_SLEEP = 100000;
+    private const DEFAULT_PERSIST_BLOCK_SIZE = 1000;
+
```
sandrokeil: @prolic There is a little issue, I don't know which option belongs to which `create*` method. For instance, `OPTION_PERSIST_BLOCK_SIZE` is only used in `createReadModelProjection()`. Maybe we should prefix the options, so you see for which method which option is available? An option without a prefix is available for all create methods?

Another solution is, the projection manager gets a factory for each type injected e.g. `QueryFactory`, `ReadModelProjectionFactory` and `ProjectionFactory` to decouple this stuff and put the options to each factory. The factory gets the event store on `__invoke`. /cc @oqq @codeliner

codeliner:
Ok I took a second look today and I'd like to continue the discussion a bit. Some thoughts:

A projection needs to poll my event-store.
The event-store is the most important piece of infrastructure I have, because the only thing that really matters is that my events are written on disk. Ideally they are replicated on a few other disks located at the other end of the world.
So I want to protect my primary event-store from any unnecessary load. The argument now is: Ok, then let the projection poll a slave/secondary. BUT I cannot write the projection state to a slave. It is a read only connection.
In mongoDB I can configure the `readPreference` when connecting to a mongo cluster, but I'd still have the problem that I would need to add load on my primary event-store just to update the projection state. My super fast append only event-store now has the second task of also updating projection state. 
I'm ok if this is the default configuration BUT I do need a way to persist my projection state somewhere else. 

@prolic 
> more interfaces, more implementations, is the projection status store another repostory? if so, more composer dependencies

Yeah, definitely a drawback. Fortunately, the ProjectionStatusStore would be as simple as the SnapshotStore and we could even provide a ProjectionStatus-Http-Service to manage all projections with a single centralized Microservice (still a few instances behind a load balancer but a company wide available service with its own authentication and authorization...)

@prolic @oqq 
> Ok now I understand why you would couple eventstore and projections.

I don't get that. Can you explain me the problem with an example?

prolic:
@sandrokeil having an option that is unused in a specific use case is not a problem, as it would get ignored, no exception is thrown, so I would leave it as is.

@codeliner @oqq projection store... Damn it, you drive me crazy :-) Will get implemented today.

prolic:
I thought about the whole thing again, I'll put my thoughts here and let's see what you have to say about it. Let's do some maths.

About event store performance, we only talk about write operations, all reads are handled by the read models. For a write operation the following is done:

1) Fetch current state from snapshots (optional)
2) Fetch missing events from last snapshot or start at event no 1
3) Append new events to the store

I tested PostgreSQL on my notebook with pgbench, it gives me approx 260 transactions per second. With two parallel running benchmarks about 150 transactions per second per benchmark. A test-transaction looks like:
```
    BEGIN;

    UPDATE pgbench_accounts SET abalance = abalance + :delta WHERE aid = :aid;

    SELECT abalance FROM pgbench_accounts WHERE aid = :aid;

    UPDATE pgbench_tellers SET tbalance = tbalance + :delta WHERE tid = :tid;

    UPDATE pgbench_branches SET bbalance = bbalance + :delta WHERE bid = :bid;

    INSERT INTO pgbench_history (tid, bid, aid, delta, mtime) VALUES (:tid, :bid, :aid, :delta, CURRENT_TIMESTAMP);

    END;
```

More information about this here: https://www.postgresql.org/docs/devel/static/pgbench.html

So we are talking here about performance on a notebook - with the right hardware you get easily a couple of thousands of transactions per second.

So what we can say so far, the event store should be able to handle 2K transactions per second or more, with the right hardware (spread across multiple php processes) - 1K for reading events to load the aggregate, 1K for appending to the stream. These are only estimates, and the performance is highly dependent on the hardware you are using. Your real numbers might be much higher or lower.

Now let's bring projections into play. From the last few applications I have build, I would say 25 running projections is a pretty high number, but let's assume we have 100. I don't think you'll ever have more than that. The projection table would have a maximum of 100 entries - every lookup should be extremely fast, the table is nearly empty.

Again, no load on the server is not the problem at all, we all agree without load there is no problem, so we are talking about very high performance applications where even PostgreSQL hits its limit from time to time, we are talking about 2k transactions per second. That means, the projections are also not idle and have to handle a lot of events. That means, the projections would probably load new events and process them every 5 seconds (we have high load, it doesn't go much faster). So we are talking about 100 projections doing stuff on DB every 5 seconds, so 20 reads per seconds (reading events for processing) and 20 writes per second (updating projection state).

Let's take this into account:
1K reads from the application + 20 reads for the projections + 1K writes from the application + 20 writes for the projections. So for our 2K transactions per second, we add 40! Does that make any difference at all? To me it seems like the projections make no difference at all on a server with high load. The performance gain of pushing the projection storage somewhere else is vanishingly small. On the other hand it adds a lot of complexity, the setup is much harder to do and there is much more to document and explain.

Seeing this, I even think that having snapshots on a different storage (like pdo-event-store with mongodb-snapshot-store) makes perhaps even no big difference for the application performance, but also increases complexity that can be reduced to a bare minimum. - A benchmark would give more insights.

My conclusions:
The event-store can handle a lot of write operations per second, depending on the used hardware.
The event-store can also handle up to 100 projections without any big performance difference compared to when having no projections at all.
The event-store could probably also handle the snapshots itself.

This would reduce the complexity a lot as you only had to setup the event store and you are ready to go with full features, without installing additional packages and additional configuration.

Last side note:
Looking at https://geteventstore.com/  - they handle the projections also on the event-store itself, without extra storage for those few projections. It's also able to handle a few thousands transactions per second - so we should be somewhere nearby - without benchmarks and pessimistic approach we are still slower by a factor of 2 or 3, but not 10 or 20.

oqq:
@prolic performance was not really an issue for me. Its more about infrastructure restrictions. I have already provide an example in chat for my current project. Also, I subscribe what @codeliner has noted about that.

@codeliner about:
> Ok now I understand why you would couple eventstore and projections.

The main reason is, that it has be possible to manage all projections without knowledge about the implementation of that projections. (e.g. stopProjection)

If every projection implements an own "ProjectionStorage", than you have to know all that implementations where you manage projections. I guess thats the reason why @prolic has stored the projections near the event store.

Every projection needs the ability the read the event store. If the ProjectionManager has to be generic, than we need either a projection store, or we handle projection management with events, stored in the event store.

So the ProjectionManager only need to know the EventStore and triggers `stopProjectionWithName`. All projections knows where they store here state (e.g. in a mongodb) and need also to read the event store. So instead of read the projections table, they could also read the projection stream events. And than set her own status. 

Or we use a `ProjectionStore`. But than there is no easy way to use more than one projection storage. E.g some projections on mongo, some on redis.

I would like it to store the projection state near the read model content. If the redis server goes down, the state is NULL and the projections automatically fills himself again.

@prolic has argued projection events are to overloaded. But I guess this would solve more issues as a ProjectionStore.

codeliner:
ok ok, now you drive me crazy @prolic @oqq :D

The maths sounds good. Maybe I'm a bit too paranoid, because I saw slow database systems too often in my developer life ;)
Question is: if the performance is still high with let's say one million events in a stream?

Anyway, we've discussed the issue long enough. I don't want to stress the topic more than necessary.
But I disagree on putting snapshots into the event-store. I really don't think that Postgres is a silver bullet, because nothing is.
I guess In-Memory snapshots are the most efficient way. Also conceptually snapshots have nothing to do with the event store.

Back to projections...
I like the projection stream idea, but I think @prolic is right here, that this is overloaded. At least if projection events are stored forever like normal domain events. Your projections stream would grow very fast but without a real business value. You are simply not interested in the history of a projection. If it fails it should write that failure into a log file and die.
But I do like the idea that a projection is an Actor (yes, Actor Model ;)).
Two projections with the same name (the same address) are not allowed to run at the same time.
Projections are persistent and they have local state.
Projections process one message at a time.

That said, if we find a way to put a mailbox in front of every projection and that mailbox has the adress of the projection name (projection name could be connected with an ip + port or similar) then we could use messages to control the projection from the outside and the projection could also produce status messages that are sent to a monitoring endpoint (yet another Actor,damn, maybe I've read too much about the Actor Model lately)

prolic:
I am plannig on providing an akka like actor to the php world. This will be highly experimental in the first development stage. But this will take time and when it will be planned together with v7, we cannot ship v7 before the end of the year. That said, we should forget about actors for now, focus on what we got and bring actors later, if changes are needed later on, then maybe in v8.

I'd say let's keep snapshots as it is now and keep the projection manager as implemented in this branch. If you need to reset the projection for whatever reason, simply ask the manager to reset your projection.

Additional projection storage makes the system too complex without much benefits.

codeliner:
I'm ok with @prolic 's suggestion.

oqq:
If we move the projection methods from event store interface to the projection manager how proposed with this PR, then I am fine with the current state to release.

prolic:
:+1:

Am 23.02.2017 16:40 schrieb "Eric Braun" <notifications@github.com>:

prolic:
Ok this one is ready - I'll prepare the changes for pdo-event-store (started already) so we can merge both together.

prolic:
@codeliner @oqq ready for final review !!!

src/InMemoryEventStore.php
```diff
@@ -102,23 +103,27 @@ public function load(
...
             /* @var Message $streamEvent */
-            if ($this->matchesMetadata($metadataMatcher, $streamEvent->metadata())
-                && ((null === $count
-                        && ($key + 1) >= $fromNumber
-                    ) || (null !== $count
-                        && ($key + 1) >= $fromNumber
-                        && ($key + 1) <= ($fromNumber + $count - 1)
-                    )
-                )
+            if (($key + 1) >= $fromNumber
```
prolic: this code should run at max speed

src/InMemoryEventStore.php
```diff
@@ -139,24 +144,28 @@ public function loadReverse(
...
             /* @var Message $streamEvent */
-            if ($this->matchesMetadata($metadataMatcher, $streamEvent->metadata())
-                && ((null === $count
-                        && ($key + 1) <= $fromNumber
-                    ) || (null !== $count
-                        && ($key + 1) <= $fromNumber
-                        && ($key + 1) >= ($fromNumber - $count + 1)
-                    )
-                )
+            if (($key + 1) <= $fromNumber
```
prolic: this code should run at max speed

src/InMemoryEventStore.php
```diff
@@ -281,18 +281,7 @@ public function fetchStreamNames(
...
-                }
-
-                if ($metadataMatcher && ! $this->matchesMetadata($metadataMatcher, $data['metadata'])) {
-                    continue;
-                }
-
-                $result[] = new StreamName($streamName);
-                ++$found;
-            } elseif (null === $filter || $filter === $streamName) {
+            if (null === $filter || $filter === $streamName) {
```
prolic: ok

src/InMemoryEventStore.php
```diff
@@ -314,16 +303,46 @@ public function fetchStreamNames(
...
-    {
-        if (null === $filter && $regex) {
-            throw new Exception\InvalidArgumentException('No regex pattern given');
+    public function fetchStreamNamesRegex(
+        string $filter,
+        ?MetadataMatcher $metadataMatcher,
+        int $limit = 20,
+        int $offset = 0
+    ): array {
+        if (false === @preg_match("/$filter/", '')) {
```
prolic: Ain't it much slower to register & unregister the error handler for that single line of code?

src/InMemoryEventStore.php
```diff
@@ -314,16 +303,46 @@ public function fetchStreamNames(
...
+        $result = [];
+
+        $skipped = 0;
+        $found = 0;
+
+        $streams = $this->streams;
+        ksort($streams);
+
+        foreach ($streams as $streamName => $data) {
+            if (! preg_match("/$filter/", $streamName)) {
```
prolic: does not work with keys

src/InMemoryEventStore.php
```diff
@@ -344,14 +363,7 @@ function (array $result, string $streamName): array {
...
         foreach ($categories as $category) {
-            if ($regex) {
-                if (! preg_match("/$filter/", $category)) {
-                    continue;
-                }
-
-                $result[] = $category;
-                ++$found;
-            } elseif (null === $filter || $filter === $category) {
+            if (null === $filter || $filter === $category) {
```
prolic: done

PR #268, opened 2017-02-20
improves fetchProjectionNames method

oqq:
Hi @prolic !

This PR
- removes regex test with error suppression
- moves the regex test to a separate method (maybe for later usage, move it back if not wanted)
- removes the ksort on an indexed array .. makes no sense here ;)
- removes the overloaded foreach loop. My approach should return the same result.
- removes else \o/

### DIFF
--- a/src/Projection/InMemoryProjectionManager.php
+++ b/src/Projection/InMemoryProjectionManager.php
@@ -112,14 +112,26 @@ final class InMemoryProjectionManager implements ProjectionManager
         throw new Exception\RuntimeException('Stopping a projection is not supported in ' . get_class($this));
     }
 
-    public function fetchProjectionNames(?string $filter, bool $regex, int $limit, int $offset): array
+    public function fetchProjectionNames(?string $filter, int $limit = 20, int $offset = 0): array
     {
-        if ($regex) {
-            return $this->fetchProjectionNamesRegex($filter, $limit);
+        if (0 >= $limit) {
+            throw new Exception\OutOfRangeException(
+                'Invalid limit "'.$limit.'" given. Must be greater than 0.'
+            );
+        }
+
+        if (0 > $offset) {
+            throw new Exception\OutOfRangeException(
+                'Invalid offset "'.$offset.'" given. Must be greater or equal than 0.'
+            );
         }
 
         if (null === $filter) {
-            return array_slice(array_keys($this->projections), $offset, $limit);
+            $result = array_keys($this->projections);
+            $result = array_slice($result, $offset, $limit);
+            sort($result, \SORT_NATURAL);
+
+            return $result;
         }
 
         if (isset($this->projections[$filter])) {
@@ -127,6 +139,36 @@ final class InMemoryProjectionManager implements ProjectionManager
         }
 
         return [];
+    }
+
+    public function fetchProjectionNamesRegex(string $regex, int $limit = 20, int $offset = 0): array
+    {
+        if (0 >= $limit) {
+            throw new Exception\OutOfRangeException(
+                'Invalid limit "'.$limit.'" given. Must be greater than 0.'
+            );
+        }
+
+        if (0 > $offset) {
+            throw new Exception\OutOfRangeException(
+                'Invalid offset "'.$offset.'" given. Must be greater or equal than 0.'
+            );
+        }
+
+        set_error_handler(function ($errorNo, $errorMsg): void {
+            throw new Exception\RuntimeException($errorMsg);
+        });
+
+        try {
+            $result = preg_grep("/$regex/", array_keys($this->projections));
+            sort($result, \SORT_NATURAL);
+
+            return array_slice($result, $offset, $limit);
+        } catch (Exception\RuntimeException $e) {
+            throw new Exception\InvalidArgumentException('Invalid regex pattern given', 0, $e);
+        } finally {
+            restore_error_handler();
+        }
     }
 
     public function fetchProjectionStatus(string $name): ProjectionStatus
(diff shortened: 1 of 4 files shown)


prolic:
Travis failed

src/Projection/InMemoryProjectionManager.php
```diff
@@ -114,46 +114,19 @@ public function stopProjection(string $name): void
 
     public function fetchProjectionNames(?string $filter, bool $regex, int $limit, int $offset): array
```
prolic: Remove regex param

src/Projection/InMemoryProjectionManager.php
```diff
@@ -192,4 +165,26 @@ public function fetchProjectionState(string $name): array
 
         return $this->projections[$name]->getState();
     }
+
+    private function fetchProjectionNamesRegex(?string $filter, int $limit): array
```
prolic: Make this public and add to interface

prolic: Remove ? from filter param

src/Projection/InMemoryProjectionManager.php
```diff
@@ -192,4 +165,26 @@ public function fetchProjectionState(string $name): array
...
+        }
+
+        set_error_handler(function($errorNo, $errorMsg): void {
+            throw new Exception\RuntimeException($errorMsg);
+        });
+
+        try {
+            $result = preg_grep("/$filter/", array_keys($this->projections));
+
+            return array_slice($result, 0, $limit);
```
prolic: This is unsorted

oqq: Your previous code was also unsorted. So I am not sure if it was expected to sort.

oqq:
Should the $offset parameter be optional and also there on regex method?

prolic:
We can make offset = 0 by default, limit = 20 maybe?

And yes, we should sort by name asc.

Am 21.02.2017 16:20 schrieb "Eric Braun" <notifications@github.com>:

oqq:
I would set $limit = NULL and test for >= 1
?

prolic:
But null is not >= 1?

Am 21.02.2017 16:28 schrieb "Eric Braun" <notifications@github.com>:

oqq:
null means no limit .. if limit is not null, then it must be >= 1

prolic:
Without limit could return 1mio results. I think that's not a good idea.

Am 21.02.2017 16:30 schrieb "Eric Braun" <notifications@github.com>:

oqq:
ah ok .. than 20

src/Projection/InMemoryProjectionManager.php
```diff
@@ -112,48 +112,63 @@ public function stopProjection(string $name): void
...
+            throw new Exception\OutOfRangeException(
+                'Invalid offset "'.$offset.'" given. Must be greater or equal than 0.'
+            );
         }
 
-        $result = [];
+        if (null === $filter) {
+            $result = array_keys($this->projections);
+            $result = array_slice($result, $offset, $limit);
+            sort($result, \SORT_NATURAL);
```
prolic: I am not sure about natural sort. I think mysql sort by name ASC is not natural, right? We should use the same sorting algorithm with all implementations, so it's consistent.

prolic: Also we need to sort first, before we slice the result, otherwise you get different results.

src/Projection/InMemoryProjectionManager.php
```diff
@@ -112,48 +112,63 @@ public function stopProjection(string $name): void
         throw new Exception\RuntimeException('Stopping a projection is not supported in ' . get_class($this));
     }
 
-    public function fetchProjectionNames(?string $filter, bool $regex, int $limit, int $offset): array
+    public function fetchProjectionNames(?string $filter, int $limit = 20, int $offset = 0): array
     {
-        if (null === $filter && $regex) {
-            throw new Exception\InvalidArgumentException('No regex pattern given');
+        if (0 >= $limit) {
```
prolic: $limit < 1

src/Projection/InMemoryProjectionManager.php
```diff
@@ -112,48 +112,63 @@ public function stopProjection(string $name): void
...
-            }
+        if (0 > $offset) {
+            throw new Exception\OutOfRangeException(
+                'Invalid offset "'.$offset.'" given. Must be greater or equal than 0.'
+            );
         }
 
-        return $result;
+        set_error_handler(function ($errorNo, $errorMsg): void {
+            throw new Exception\RuntimeException($errorMsg);
```
prolic: Is the error message meaningful enough? With this: `preg_match('/invalid)/', '');` I receive the message: `Compilation failed: unmatched parentheses at offset 7` - Maybe it should be prepended with "Invalid regex: " ?

src/Projection/InMemoryProjectionManager.php
```diff
@@ -112,48 +112,63 @@ public function stopProjection(string $name): void
...
+        });
+
+        try {
+            $result = preg_grep("/$regex/", array_keys($this->projections));
+            sort($result, \SORT_NATURAL);
```
prolic: first sort, than slice, very good! Again same concern about natural sorting.

oqq:
If have changes made and added another unit test for the string filter usage.
I am not sure if that makes sense, since it would always response only one stream. So limit and offset is unnecessary for that use case.

Also i am not sure if it makes sense to throw a "no projection found" exception instead of returning an empty array.

prolic:
thanks!

Issue #273, opened 2017-03-10
Rename projection to projector


prolic:
A projector creates a projection, right? We should think about renaming the classes where applicable.

Thoughts?

@codeliner @oqq @basz

basz:
I agree; The result of a projector would be a projection

If you take this further a ReadModel could also be called a Projection...

prolic:
@basz I think this could let to confusion, if you need to inject a projection into a projector in order to create a projection. Inject a read model implementation into a projector to create a projection sounds way better to me.

prolic:
A pull-request is created, so we can discuss changes also there and see what it looks like.

PR #275, opened 2017-03-10
rename projection => projector where applicable

prolic:
move constants from projection manager to projector interfaces

resolves https://github.com/prooph/event-store/issues/273

Please check the new naming and check for consistency where it's still called projection, and where it's now called projector, f.e.

```
ProjectionManager::createProjection(): Projector
```

/cc @basz @oqq @bweston92

If accepted, we need to update pdo-event-store before merging, then we are ready for next beta I guess

### DIFF
--- a/src/Projection/InMemoryEventStoreProjector.php
+++ b/src/Projection/InMemoryEventStoreProjector.php
@@ -24,7 +24,7 @@ use Prooph\EventStore\Stream;
 use Prooph\EventStore\StreamName;
 use Prooph\EventStore\Util\ArrayCache;
 
-final class InMemoryEventStoreProjection implements Projection
+final class InMemoryEventStoreProjector implements Projector
 {
     /**
      * @var string
@@ -128,10 +128,10 @@ final class InMemoryEventStoreProjection implements Projection
         $this->innerEventStore = $eventStore;
     }
 
-    public function init(Closure $callback): Projection
+    public function init(Closure $callback): Projector
     {
         if (null !== $this->initCallback) {
-            throw new Exception\RuntimeException('Projection already initialized');
+            throw new Exception\RuntimeException('Projector already initialized');
         }
 
         $callback = Closure::bind($callback, $this->createHandlerContext($this->currentStreamName));
@@ -147,7 +147,7 @@ final class InMemoryEventStoreProjection implements Projection
         return $this;
     }
 
-    public function fromStream(string $streamName): Projection
+    public function fromStream(string $streamName): Projector
     {
         if (null !== $this->query) {
             throw new Exception\RuntimeException('From was already called');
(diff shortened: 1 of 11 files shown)


prolic:
@oqq @codeliner are you okay with this?

prolic:
now rebased

oqq:
Oh yes!
But I am not sure about the method names in the manager. Should we stay on method names `projection` here, or also rename they to `projector`? e.g createProjection vs createProjector.

prolic:
@oqq about createProjection VS createProjector - I thought the same thing and I think it's a hard decision here. I left it as is, because I thought it is a better naming. I want to create a projection, receive a projector, configure and run it. So that's why I didn't change it - but if you think otherwise, I am open for discussions here.

oqq:
Same thoughts about that here. Maybe a mix of booth would match. createProjector, but fetchProjectionStatus and deleteProjection. But no hard preferences from me.

prolic:
other opinions @bweston92 @codeliner @basz ?

codeliner:
>  I thought it is a better naming. I want to create a projection, receive a projector, configure and run it. 

sounds good

basz:
Geeh, surprisingly hard to get to a conclusion

> Projection has various meanings, but what they all have in common is that something is sent out or forward. A movie is projected onto a screen; a skilled actress projects her voice out into a large theater without seeming to shout; and something sticking out from a wall can be called a projection.

We are 'projecting' (a process) historic events to yield a new state (a projection of state) To do this you'll need a 'projector' and this resulting state may be called a 'projection'. 

both the projector and the projection have a status of their own.

projection status
- up-to-date
- behind

projector status
- resetting
- idle (which is actually projecting while up-to-date)
- running (which is actually projecting when behind)
- etc

so;I think I would settle for;

createProjector
resetProjector (just happens to also change projection status)
fetchProjectorStatus ([these](https://github.com/prooph/event-store/blob/5ec6e8d9b41f2d5e7fbc781e1b25e5832833d083/src/Projection/InMemoryProjectionManager.php#L171) are strange)
stopProjector
createReadModelProjector
deleteProjection 
fetchProjectionNames

prolic:
@basz Why "deleteProjection" but "resetProjector"? They are very similar, delete drops and forgets about it, while reset drops and starts from scratch. So one both should be name projection or projector, not mixed.

About Projector- vs ProjectionStatus:

Well it's pretty hard to tell, whether or not a projection is behind or up-to-date, you could query same internal stuff to get an idea whether or not it's far behind, but being "up-to-date" can already be an old information, when tons of new events are coming in the next second.

Statuses like "running", "resetting", "idle" and so on can be used for projectors, but you can also say that these projections are running / idle / resetting - no big difference.

About https://github.com/prooph/event-store/blob/5ec6e8d9b41f2d5e7fbc781e1b25e5832833d083/src/Projection/InMemoryProjectionManager.php#L171: Agree, we still need to update some variables.

I'd suggest we leave it as is and clean up some variable naming, where still needed (@oqq did a great job to do this already).

Does anyone have hard reasons against the current naming?

PR #276, opened 2017-03-10
updates unit tests to reflect current implementation

oqq:
I have replaces PHP_INT_MAX by null to test real usages for current implementation.

Also I have removed some skipped tests which should only overwrite tests from extended test case, which not longer there.

And small coverage increase.


prolic:
thank you

Issue #191, opened 2016-10-24
RFC some constraints on metadata and event stream names


prolic:
I want to propose some changes here:

1) Move aggregate_type and aggregate_id to event metadata. This is really only needed in special case that one stream per aggregate is not used. With that change we can have a unified event stream structure independent of which stream strategy is used.

2) All internal metadata should be prefixed - I propose the underscore `_` for this. This means all internal metadata (currently this would be: causation_id, causation_name, aggregate_type and aggregate_id) will be prefixed with underscore. Userland metadata is forbidden to use the underscore as first character of their metadata as it's reserved for internal usage. I don't know yet if an exception is necessary or if a note in the docs to strongly discourage it is enough.

3) All internal streams like for stream-projections and so on (see: https://github.com/prooph/event-store/issues/193) will be prefixed with an underscore `_`, too. All userland streams are required not be begin with underscore as this is reserved for internal useage. Again: I don't know yet if an exception is necessary or if a note in the docs to strongly discourage it is enough.

codeliner:
I'd go for the note in the docs, otherwise we would need to check each metadata key. People should be familiar with conventions because frameworks use this a lot to make things easier.

I like the idea but we need to provide migration paths. Upgrading to the new major versions will be work and people need to be aware of that.

prolic:
I am thinking of providing a migration script for the database using the event store api, so it should work for all adapters without providing special migration scripts for every database vendor.

sandrokeil:
Sounds really good and make also things easier for new people, because of less necessary fields.

PR #280, opened 2017-03-24
Docs

prolic:
resolves https://github.com/prooph/event-store/issues/191

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -1,0 +1,63 @@
+# Migration from v6 to v7
+
+## Moved classes
+
+The `Prooph\Snapshot\*` classes are now moved to its own repository [SnapshotStore](https://github.com/prooph/snapshot-store).
+
+The `Prooph\Aggregate\*` classes are now moved into the [event-sourcing](https://github.com/prooph/event-sourcing/) repository.
+
+## Interfaces VS adapters
+
+The event store now ships with a `Prooph\EventStore\ReadOnlyEventStore` and a `Prooph\EventStore\EventStore` interface.
+No event store adapters exist any more, instead there are different implementations of an event store.
+
+## ActionEventEmitterEventStore
+
+In order to use action events like in v6, you need to wrap your event store.
+
+```php
+$eventStore = new ActionEventEmitterEventStore($eventStore);
+```
+
+or
+
+```php
+$eventStore = new TransactionalActionEventEmitterEventStore($eventStore);
+```
+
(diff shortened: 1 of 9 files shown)


docs/migration.md
```diff
@@ -0,0 +1,63 @@
...
+
+If you are using the event-store together with the [event-sourcing](https://github.com/prooph/event-sourcing/) component,
+most stuff is pretty much unchanged for you, as you don't interact with the event store directly (this is done by the
+event-sourcing component).
+
+If you are making calls to the event store yourself, take a look at the event_store docs on how the new usage is.
+
+## DB migrations
+
+If you are using v6 with MySQL (using doctrine adpater) and you want to switch to v7 with MySQL (using pdo-event-store),
```
basz: is this only true for mysql?

docs/migration.md
```diff
@@ -0,0 +1,63 @@
...
+you need to upgrade your database before you do this. The way events are persisted have changed and you cannot simply
+update your source code to make this change. You need to write a migration script, take the database offline,
+perform the migration and go back online.
+
+As this is a very tough job, we don't provide any migration script currently and for some applications a downtime is not
```
basz: if you can't go down you can't upgrade... ok noted... but if i can go down, what would I need to do? loop through table with events, create new stream, read old and persist new events. more..?

docs/projections.md
```diff
@@ -0,0 +1,170 @@
+# Projections
+
+New in v7 are queries and projectons.
+
```
basz: Reading this document I've come to realise that a query is the actually the same thing as a projection but without persistence of state.... Usually one might think of a query as a representation of (pre) calculated state. I thought for example that queries are a way to retrieve persisted projection state.

I think it would be good to add a small section here explaining this subtlety. Looping over a (sub)set of events (a stream) and updating a running state while you do so. Projections stores state, query executed each time.

prolic:
thanks for your feedback @basz - I added some more lines to the docs.

prolic:
@mablae @oqq @bweston92 please leave some feedback, if the docs help you out to understand what's going on here, tell me it needs improvements, and so on.
If possible, create a PR to my branch.

prolic:
@oqq @codeliner @basz small interface change done here. I added the locktimeout option to the interface, it will be removed from the pdo-event-store.

codeliner:
Very nice and clean explanation of the migration steps and the new projection features. Well done! If no further updates are planed it can be merged.

prolic:
thanks for the review @codeliner and @basz

PR #281, opened 2017-03-26
conflict with php 7.1.3

prolic:
@sandrokeil like this?


prolic:
I think it's not needed here, only in pdo-event-store.

PR #282, opened 2017-03-26
Disallow PHP 7.1.3 because of Datetime bug (close #279)

### LINKED ISSUE #279: Data too long for column 'locked_until'
I get this error on my snapshot projection procosses

This happened after I upgrade PHP

```
[Prooph\EventStore\Exception\RuntimeException]                
  Error 22001. Maybe the projection table is not setup?         
  Error-Info: Data too long for column 'locked_until' at row 1  
```
the locked_untill file is NULL

sandrokeil:
I've used conflict section because I think it's more verbose.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,8 @@
         "sandrokeil/interop-config": "For usage of provided factories"
     },
     "conflict": {
-        "sandrokeil/interop-config": "<2.0.1"
+        "sandrokeil/interop-config": "<2.0.1",
+        "php": "7.1.3"
     },
     "autoload": {
         "psr-4": {


sandrokeil:
Ok, we have to disable PHP 7.1.3 for Travis.

@basz I guess it's only a problem for PDO or is a better fix to increase the length of the `locked_until` field? There is no index for this field, so it could also be a varchar?

prolic:
Thanks @sandrokeil, but I think it's not needed here, only at pdo-event-store, I prepared already a PR here: https://github.com/prooph/pdo-event-store/pull/70

sandrokeil:
Agree

Issue #284, opened 2017-04-29
Support message property filters


codeliner:
It is only possible to query event streams by metadata using a `MetadataMatcher`: https://github.com/prooph/event-store/blob/master/src/ReadOnlyEventStore.php#L28

Would be nice if we could also use the other message properties as filters like
-  `uuid` to look up a specific event
- `created_at` to look up newer or older events
- `message_name` to only get events of a certain type

I can think of two ways to enable such filters:

1) Add a new `fieldType` argument to `MetadataMatcher::withMetadataMatch` method that defaults to `FIELD_TYPE::METADATA`, but can also be set to `FIELD_TYPE::MESSAGE_PROPERTY`

2) Add a new `MessagePropertyMatcher` as optional argument to `EventStore::load` and `EventStore::loadReverse`

codeliner:
My favorite is the first one. The other message properties don't belong to the metadata but they are some kind of metadata. Furthermore this option simplifies implementation in the event-store-http-api and mgmt-ui ;)

prolic:
I have no strong opinion on that yet, but here is what comes to my mind:

- uuid to look up a specific event - you have know the event stream in order to get the event by uuid. I would recommend using an event stream projection for this, so it would it make very easy to query without knowing the exact event stream

- created_at to look up newer or older events - if that's important for your analyses, put it in the metadata as well?

- message_name to only get events of a certain type - see https://github.com/prooph/standard-projections/blob/master/src/MessageNameStreamProjectionRunner.php

codeliner:
I'm looking from the event-store-mgmt-UI POV:
I'd like to provide a useful UI without forcing the user to create specific projections. Just connect to a evnet-store-http-api and you're ready to go.

We often need to connect to our production event store to lookup events and that is pain because it is hosted on AWS and you need to take some steps to get there. The lookups are requested by the business and are often related to identify problems with the defined business process or because we received wrong data from the source systems.
Int the future I'd like to have a event-store-mgmt-ui with read-only access deployed to production and connected with our authentication and authorization service. Ideally the mgmt-UI is so easy to use that I can grant access to the business people and they can lookup things by oneself. 

My request / suggestion is based on this experience. Metadata alone does not offer enough query possibilities to find information quickly. Working with plain SQL or in our case mongo queries is pain and dangerous on a production system! A dedicated but powerful query API with UI support is a must-have IMHO and it is relativ easy to add without breaking BC.

[...]

PR #286, opened 2017-05-21
Support message property filters + new operators

prolic:
- add new metadata matcher operators
- add message property matching

resolves https://github.com/prooph/event-store/issues/284

pdo-event-store updates needed before release


codeliner:
nice addition :+1:

PR #288, opened 2017-05-22
better validation in metadata matcher

prolic:


src/Metadata/MetadataMatcher.php
```diff
@@ -35,12 +35,6 @@ public function withMetadataMatch(
             $fieldType = FieldType::METADATA();
         }
 
-        if ($fieldType->is(FieldType::MESSAGE_PROPERTY())
-            && ! in_array($field, ['uuid', 'createdAt', 'created_at', 'messageName', 'message_name'], true)
-        ) {
-            throw new InvalidArgumentException(sprintf('Invalid message property "%s" given', $field));
```
prolic: I needed to remove that, because the fields have different names in pdo-event-store tables

codeliner: but pdo-event-store can map them. I'd like to provide default filters with the event-store-mgmt-ui and I don't know which event-store implementation is used.

prolic:
Problem is persistence strategy can name to fields like it wants to and
interface is final.

codeliner:
ok then we need to find another way for the mgmt-ui

PR #293, opened 2017-06-30
Add projection option constants to trigger pcntl_signal_dispatch 

fritz-gerneth:
Moving these constants to the generic interface as discussed [here ](https://github.com/prooph/pdo-event-store/pull/96). Will update the other PR with the new minimum version constraint once this has been merged and released as a new version.

### DIFF
--- a/docs/projections.md
+++ b/docs/projections.md
@@ -145,6 +145,22 @@ OPTION_LOCK_TIMEOUT_MS = 'lock_timeout_ms'
 Indicates the time (in microseconds) the projector is locked. During this time no other projector with the same name can
 be started. A running projector will update the lock timeout on every loop.
 
+OPTION_PCNTL_DISPATCH = 'false'
+
+Enable dispatching of process signals to registered [signal handlers](http://php.net/manual/en/function.pcntl-signal.php) while
+the projection is running. You must still register your own signal handler and take according action.
+For example to gracefully stop the projection you could do
+```
+$projection = $projectionManager->createProjection(
+    'test_projection',
+    [ Projector::OPTION_PCNTL_DISPATCH => true, ]
+);
+pcntl_signal(SIGQUIT, function () use ($projection) {
+    $projection->stop();
+});
+$projection->run();
+```
+
 ## Read Model Projections
 
 Projections can also be used to create read models. A read model has to implement `Prooph\EventStore\Projection\ReadModel`.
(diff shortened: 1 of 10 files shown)


prolic:
Same change is missing in ReadModelProjector interface.

Do you think it's worth adding this feature also to InMemoryEventStore?

prolic:
Travis failed.

prolic:
```
1) ProophTest\EventStore\Projection\InMemoryEventStoreProjectorTest::it_throws_exception_when_unknown_event_store_instance_passed

Failed asserting that exception of type "TypeError" matches expected exception "Prooph\EventStore\Exception\InvalidArgumentException". Message was: "Argument 5 passed to Prooph\EventStore\Projection\InMemoryEventStoreProjector::__construct() must be of the type boolean, integer given, called in /home/travis/build/prooph/event-store/tests/Projection/InMemoryEventStoreProjectorTest.php on line 121" at

/home/travis/build/prooph/event-store/src/Projection/InMemoryEventStoreProjector.php:109

/home/travis/build/prooph/event-store/tests/Projection/InMemoryEventStoreProjectorTest.php:121
```

fritz-gerneth:
Not entirely sure why the test fails for travis. It's the same error I get over [here ](https://github.com/prooph/pdo-event-store/pull/96) too.  If you have any idea what might be causes this .. :)

I thought about moving the tests to the abstract classes - but then we cannot profit from test inheritence like we do in the test files (let the extending class define the projection manager, setup is outsourced to the separate files instead).

prolic:
some php-cs isses, and missing `declare(strict_types=1);`

prolic:
Will review again tomorrow and I check the tests and what's going on with travis.

prolic:
@fritz-gerneth I fixed the tests and made some small changes. Please check this.

Why exec before the command?

From php.net:
Because the way proc_open() actually works on Unix/Linux is by starting "/bin/sh -c usercmd userargs...", e.g., "/bin/sh -c /usr/bin/compress /tmp/foo".[Note 1]  That means normally your command is the child of the shell, so the pid you retrieve with proc_get_status() is the pid of the shell (PHP's child), and you have to fumble around trying to find the pid of your command (PHP's grandchild).  But if you put "exec" in front of your command, you tell the shell to *replace itself* with your command without starting another process (technically, to exec your command without forking first).  That means your command will inherit the pid of the shell, which is the pid that proc_get_status() returns.

fritz-gerneth:
I see, weird it worked on my Linux then :) (CentOs7). THanks for digging into this (y)

PR #304, opened 2017-10-14
Restructure docs

codeliner:
According to https://github.com/prooph/proophessor/issues/38#issuecomment-336200542

- shorter headlines
- introduction instead of using entire README (better overview while browsing through the docs)


codeliner:
@prooph/components-core I merge directly so that I can test the changes.

PR #308, opened 2017-10-19
Introducing NonTransactionalInMemoryEventStore

tomcizek:
Solves incompatibility of InMemoryEventStore and PdoEventStore as explained in this SO question: https://stackoverflow.com/questions/46769347/how-to-solve-incompatibility-of-inmemoryeventstore-and-pdoeventstore-in-prooph/46806028#46806028


sandrokeil:
@tomcizek Can you run `composer cs-fix` please and commit the changes. Then Travis should be happy. ;-)

src/Container/AbstractInMemoryEventStoreFactory.php
```diff
@@ -0,0 +1,169 @@
...
+        if ($config['read_only']) {
+            $eventStore = new ReadOnlyEventStoreWrapper($eventStore);
+        }
+
+        if (! $config['wrap_action_event_emitter']) {
+            return $eventStore;
+        }
+
+        if (! isset($config['event_emitter'])) {
+            $eventEmitter = new ProophActionEventEmitter(TransactionalActionEventEmitterEventStore::ALL_EVENTS);
```
prolic: you attach all events for non-transactional event-store

src/Container/InMemoryEventStoreFactory.php
```diff
@@ -12,149 +12,26 @@
...
-                throw ConfigurationException::configurationError(sprintf(
-                    'Plugin %s does not implement the Plugin interface',
-                    $pluginAlias
-                ));
-            }
+    protected function createActionEventEmitterDecorator(
+        EventStore $eventStore,
+        ActionEventEmitter $actionEventEmitter
+    ): ActionEventEmitterEventStore {
+        assert($eventStore instanceof TransactionalEventStore);
```
prolic: don't use assert function, use if not and throw exception

tomcizek: What is the reason for that? The intention of this assert is just down-type EventStore because in same class I know for sure TransactionalEventStore will be there. The only way it can fail here is to modify this vendor class which would result in failing test. Maybe adding some description for eventual maintainer of this class, in case he would accidentaly not return TransactionalEventStore in createEventStoreInstance() method would help? 
Adding Exception would result in untestable and uncovered code at the line it would be thrown or maybe it would take some ugly magic to do that.

tomcizek:
Yes, seems reasonable, I will do it this week. :)

docs/interop_factories.md
```diff
@@ -60,6 +60,10 @@ If the requirements are met, you just need to add a new section in your applicat
 
 $eventStore = $container->get('inmemoryeventstore');
 
+By default, `InMemoryEventStore` which is transactional, is created by factory. If you want to change this behaviour to 
+create `NonTransactionalInMemoryEventStore`, simply override default config of given event store to 
+`'transactional' => true`
```
prolic: you mean false?

src/Container/InMemoryEventStoreFactory.php
```diff
@@ -154,7 +163,45 @@ public function defaultOptions(): iterable
...
+        if ($this->isTransactional($config)) {
+            return TransactionalActionEventEmitterEventStore::ALL_EVENTS;
+        }
+
+        return ActionEventEmitterEventStore::ALL_EVENTS;
+    }
+
+    private function createEventStore(array $config): EventStore
+    {
+        if ($this->isTransactional($config)) {
```
prolic: we should do this outside of this method, so the same method is not called 3 times, when only 1 call is needed

src/Container/InMemoryEventStoreFactory.php
```diff
@@ -154,7 +163,45 @@ public function defaultOptions(): iterable
...
             'plugins' => [],
             'wrap_action_event_emitter' => true,
+            'transactional' => true,
             'read_only' => false,
         ];
     }
+
+    private function determineEventsForDefaultEmitter(array $config): array
+    {
+        if ($this->isTransactional($config)) {
```
prolic: same here

src/Container/InMemoryEventStoreFactory.php
```diff
@@ -154,7 +163,45 @@ public function defaultOptions(): iterable
...
+
+        return new NonTransactionalInMemoryEventStore();
+    }
+
+    private function createActionEventEmitterDecorator(
+        array $config,
+        EventStore $eventStore,
+        ActionEventEmitter $actionEventEmitter
+    ): ActionEventEmitterEventStore {
+        if ($this->isTransactional($config)) {
```
prolic: again

src/InMemoryEventStore.php
```diff
@@ -30,17 +30,17 @@
     /**
      * @var array
      */
-    private $streams = [];
+    protected $streams = [];
```
prolic: why is this changed to protected? it's a final class anyway

src/InMemoryEventStore.php
```diff
@@ -333,7 +269,6 @@ public function fetchStreamNamesRegex(
 
         $result = [];
 
-        $skipped = 0;
```
prolic: why is this removed?

src/InMemoryEventStore.php
```diff
@@ -555,4 +490,68 @@ private function match(Operator $operator, $value, $expected): bool
...
+    {
+        return $this->inTransaction;
+    }
+
+    /**
+     * @throws \Exception
+     *
+     * @return mixed
+     */
+    public function transactional(callable $callable)
```
prolic: I see no useful changes in this class, can you revert the changes please?

src/NonTransactionalInMemoryEventStore.php
```diff
@@ -0,0 +1,480 @@
...
+use Prooph\EventStore\Metadata\MetadataMatcher;
+use Prooph\EventStore\Metadata\Operator;
+use Prooph\EventStore\Util\Assertion;
+
+final class NonTransactionalInMemoryEventStore implements EventStore
+{
+    /**
+     * @var array
+     */
+    protected $streams = [];
```
prolic: why protected?

src/Container/InMemoryEventStoreFactory.php
```diff
@@ -154,7 +170,44 @@ public function defaultOptions(): iterable
...
+
+        return new NonTransactionalInMemoryEventStore();
+    }
+
+    private function createActionEventEmitterDecorator(
+        EventStore $eventStore,
+        ActionEventEmitter $actionEventEmitter
+    ): ActionEventEmitterEventStore {
+        if ($this->isTransactional) {
+            assert($eventStore instanceof TransactionalEventStore);
```
prolic: actually, the TransactionalActionEventEmitterEventStore expects TransactionalEventStore as argument in the constructor, so this can be removed as well

src/Container/InMemoryEventStoreFactory.php
```diff
@@ -198,8 +197,7 @@ private function createActionEventEmitterDecorator(
         ActionEventEmitter $actionEventEmitter
     ): ActionEventEmitterEventStore {
         if ($this->isTransactional) {
-            assert($eventStore instanceof TransactionalEventStore);
-
+            /** @var TransactionalActionEventEmitterEventStore $eventStore */
```
prolic: TransactionalActionEventEmitterEventStore is wrong, it's TransactionalEventStore, but you can also remove this line

tomcizek: Ah sorry, did that in hurry. Without that line IDE shines with error (and autosuggest does not work) which I don't like... Is this line problem or can I leave it after fixing classname?

prolic: sure, leave but fix class name

prolic:
Thanks a lot !!!

PR #320, opened 2018-02-06
fixed metadata enum class doc-blocks

ged15:
Fixed metadata enum class doc-blocks according to https://docs.phpdoc.org/references/phpdoc/tags/method.html.
Super minor change, but breaks PHPStan analysis if these classes are used in project.


prolic:
Thanks!

PR #331, opened 2018-04-07
Add projection option update lock threshold

codeliner:
See updated docs for reasons.
Default value of the new projection option is 0, so existing projections are not effected by this new optional config param.


prolic:
pdo-event-store implementation is coming?

codeliner:
yes, that's the precondition for pdo-event-store

PR #338, opened 2018-04-26
Dispatch PCNTL signal after each event for immediately shutdown

### LINKED ISSUE #337: pcntl_signal_dispatch should be called after each event was processed
If we take a look at the [PdoEventStoreReadModelProjector](https://github.com/prooph/pdo-event-store/blob/master/src/Projection/PdoEventStoreReadModelProjector.php#L446) the PCNTL signal is only dispatched if all events of all streams was processed. 

Assume we have 10.000 events and one event takes `0.1s` for processing and you want to graceful shutdown the projection. The projection stops only after all events of all streams was processed (SIGNAL or `stop()`).

The `pcntl_signal_dispatch()` should be called after each event was processed, then the interruption is used immediately (depending on the duration of processing one event). And by calling `stop()`, the state should also be persisted to not loose already processed events.

Am I right?

sandrokeil:
close #337

I have to upgrade PHPUnit, otherwise the small test timeout was not working. If you remove the additional `pcntl_signal_dispatch();` PHPUnit will fail with risky tests.


prolic:
What about non read model projectors and queries?

sandrokeil:
@prolic Good point, will do that too in the next days.

sandrokeil:
This is ready for review.

codeliner:
LGTM

PR #343, opened 2018-06-28
Add support to read transactional data in InMemoryEventStore (close #342)

### LINKED ISSUE #342: InMemoryEventStore can't read it's own events if in transaction
The `InMemoryEventStore` can't read it's own events if in transaction but I think it's a wrong behaviour. Normally if you are in a transaction, you can read your own stuff which was written to database in this transaction session.

It exists also a [test](https://github.com/prooph/event-store/blob/master/src/InMemoryEventStore.php#L50-L54) which checks if a stream already exists in transaction.

I will bring a PR which writes the transaction events also to `streams` property and use the `cachedStreams` only to remove certain events on roll back from `streams`.

Any concerns about this?

sandrokeil:
close #342

### DIFF
--- a/tests/TransactionalActionEventEmitterEventStoreTest.php
+++ b/tests/TransactionalActionEventEmitterEventStoreTest.php
@@ -55,7 +55,7 @@ class TransactionalActionEventEmitterEventStoreTest extends TestCase
 
         $this->eventStore->create($stream->reveal());
 
-        $this->assertFalse($this->eventStore->hasStream($streamName));
+        $this->assertTrue($this->eventStore->hasStream($streamName), 'It can not read transactional data');
 
         $this->eventStore->commit();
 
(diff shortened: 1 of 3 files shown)


tests/TransactionalActionEventEmitterEventStoreTest.php
```diff
@@ -55,7 +55,7 @@ public function it_works_transactional(): void
 
         $this->eventStore->create($stream->reveal());
 
-        $this->assertFalse($this->eventStore->hasStream($streamName));
+        $this->assertTrue($this->eventStore->hasStream($streamName), 'It can not read transactional data');
```
sandrokeil: Does this affect other repositories?

tests/TransactionalActionEventEmitterEventStoreTest.php
```diff
@@ -82,7 +82,7 @@ public function it_rolls_back_transaction(): void
...
```
sandrokeil: Does this affect other repositories?

sandrokeil:
We need a `InMemoryConnection` class to get transaction support for all `InMemory*` classes to emulate it like a PDO connection.

tests/TransactionalActionEventEmitterEventStoreTest.php
prolic: Yes it does

PR #346, opened 2018-09-14
[WIP] feature/countable stream iterator

basz:
@prolic This is what I came up with. As always naming is an issue, php isn't consistent when it comes to interface naming. IMO 'Iterator' should have been called 'Iterable'.

This PR;
- introduces an StreamIterator namespace.
- provides an abstract StreamIterator class. It `implements` Iterator and Countable.
- implements an EmptyStreamIterator and an ArrayStreamIterator from the abstract.
- changes the EventStore implementations load and loadReverse methods to return a StreamIterator implementation. It does this without changing the return type of `Iterator`. It does not seem not possible to change the `Prooph\EventStore\ReadOnlyEventStore` interface to have a return type of `StreamIterator` without a BC break. :-( I do believe it is a minor BC as 'probably' will only break tests that use ArrayIterator or EmptyIterator (like you did).

Q: Is there a reason you are utilizing the EmptyIterator over an empty ArrayIterator?

Although I also provided an EmptyStreamIterator (you were using EmptyIterator) I think an ArrayStreamIterator (with an empty array) would be just as good. Both are countable now, but it might be a possibly BC break. Probably anyone currently checking for `$i instanceof EmptyIterator` somewhere.


prolic (changes requested):
looks good.

prolic (approved):

PR #350, opened 2018-11-01
Update fromStream for Projection, Query and ReadModelProjection

fjogeleit:
This change make it possible to query for a single aggregate with a SingleStream- and SimpleStreamStrategy behind.

### DIFF
--- a/src/Projection/InMemoryEventStoreQuery.php
+++ b/src/Projection/InMemoryEventStoreQuery.php
@@ -19,6 +19,7 @@ use Prooph\EventStore\EventStore;
 use Prooph\EventStore\EventStoreDecorator;
 use Prooph\EventStore\Exception;
 use Prooph\EventStore\InMemoryEventStore;
+use Prooph\EventStore\Metadata\MetadataMatcher;
 use Prooph\EventStore\NonTransactionalInMemoryEventStore;
 use Prooph\EventStore\StreamName;
 
@@ -79,6 +80,11 @@ final class InMemoryEventStoreQuery implements Query
      */
     private $triggerPcntlSignalDispatch;
 
+    /**
+     * @var MetadataMatcher|null
+     */
+    private $metadataMatcher;
+
     public function __construct(EventStore $eventStore, bool $triggerPcntlSignalDispatch = false)
     {
         $this->eventStore = $eventStore;
@@ -119,13 +125,14 @@ final class InMemoryEventStoreQuery implements Query
         return $this;
     }
 
-    public function fromStream(string $streamName): Query
+    public function fromStream(string $streamName, MetadataMatcher $metadataMatcher = null): Query
     {
         if (null !== $this->query) {
             throw new Exception\RuntimeException('From was already called');
         }
 
         $this->query['streams'][] = $streamName;
+        $this->metadataMatcher = $metadataMatcher;
 
         return $this;
     }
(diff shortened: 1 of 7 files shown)


src/Projection/InMemoryEventStoreQuery.php
```diff
@@ -396,7 +414,7 @@ private function prepareStreamPositions(): void
         }
 
         // stream names given
-        foreach ($this->query['streams'] as $stream) {
+        foreach ($this->query['streams'] as $key => $stream) {
```
gquemener: `$key` is unused and should be ignored, shouldn't it?

src/Projection/Query.php
```diff
@@ -29,6 +30,8 @@ public function fromStream(string $streamName): Query;
 
     public function fromStreams(string ...$streamNames): Query;
 
+    public function fromStreamWithMetadataMatcher(string $streamName, MetadataMatcher $metadataMatcher): Query;
```
gquemener: This is a BC break and should be released in a new major version, shouldn't it?

src/Projection/InMemoryEventStoreQuery.php
fjogeleit: yes, your right.

src/Projection/InMemoryEventStoreQuery.php
```diff
@@ -79,6 +80,11 @@
      */
     private $triggerPcntlSignalDispatch;
 
+    /**
+     * @var MetadataMatcher
```
prolic: MetadataMatcher|null

src/Projection/Query.php
prolic: Exactly, my bad. I think this should work:
`public function fromStream(string $streamName, MetadataMatcher $metadataMatcher = null): Query;`

src/Projection/InMemoryEventStoreQuery.php
fjogeleit: fixed

src/Projection/ReadModelProjector.php
```diff
@@ -35,7 +36,7 @@
      */
     public function init(Closure $callback): ReadModelProjector;
 
-    public function fromStream(string $streamName): ReadModelProjector;
+    public function fromStream(string $streamName, MetadataMatcher $metadataMatcher = null): ReadModelProjector;
```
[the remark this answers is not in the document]
fjogeleit: I removed it from the interfaces

prolic:
Looks good so far. Before merging this, I'd like to see two things.
1) implementation for pdo event store
2) docs

fjogeleit:
A PR for the PDO Event Store is already open:  https://github.com/prooph/pdo-event-store/pull/177

Docs will be updated asap

src/Projection/ReadModelProjector.php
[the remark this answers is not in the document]
gquemener: Changing a method signature (even with optional parameters) is a BC break and all custom implementations will break after this change.

```
php > interface X
php > { 
php { public function foo(string $a, int $b = null): void
php { ;
php { }
php > class A implements X
php > {
php { public function foo(string $a): void
php { {
php { }
php { }

Fatal error: Declaration of A::foo(string $a): void must be compatible with X::foo(string $a, ?int $b = NULL): void in php shell code on line 1
```

A BC solution would have been to create a separate interface, which would have provided the `fromStreamWithMetadataMatcher` method OR to release this in a new major version.

prolic:
The change is only in implementation, not in the interface itself.

gquemener:
Oh right, I see.

Sorry, I got confused with the github discussion feed.

prolic:
No problem

PR #363, opened 2019-03-24
Add Support for "event_name" field for MetadataMatcher in InMemoryEventStore

fjogeleit:
I have a `Prooph\EventStore\Projection\Query` wo is using a MetadataMatcher for the MessageProperty `event_name` (Using PostgreSQL EventStore`.

When I try to test this Query local with a InMemoryEventStore this MetadataMatcher is not working because this field is only supported as `messageName` / `message_name`.

To make this test possible this PR also add support for  the `event_name` as field name in the InMemoryEventStore

### DIFF
--- a/examples/event/QuickStartSucceeded.php
+++ b/examples/event/QuickStartSucceeded.php
@@ -2,8 +2,8 @@
 
 /**
  * This file is part of prooph/event-store.
- * (c) 2014-2018 prooph software GmbH <contact@prooph.de>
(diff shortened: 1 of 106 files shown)


fjogeleit:
The second commit is only for updating the comment section to pass the Travis CI checks

prolic:
Why do you refuse to use `message_name`?

fjogeleit:
Because `message_name` doensn't work in productive for a PDO based query. It's generating a `where` clause `message_name = :value` for example but the field name in the PDO EventStore is called `event_name`

PR #364, opened 2019-03-26
Improve DateTime::create()

enumag:
```
TypeError(code: 0): Return value of Prooph\\EventStore\\Util\\DateTime::create() must be an instance of DateTimeImmutable, bool returned at /usr/app/vendor/prooph/event-store/src/Util/DateTime.php:31
```

I know this should not happen but it did happen for me. Unfortunately it's not easy to reproduce and I don't have the string nor the stack trace. This should help me debug any future cases.


enumag:
(Do a squash-merge pls.)

PR #367, opened 2019-05-13
Improvement/merged stream iterator

basz:
implementation of https://github.com/prooph/pdo-event-store/issues/201

some notes;
- works like any other iterator (no array result from current with results from multiple iterators)
- can work with any number of streams (but not zero)
- can query for streamName of `current` message (needed for tracking of projection positions)
- invalid iterators may become valid again while still iterating other valid ones without problem
- think it should be still fastish

### DIFF
--- a/src/Projection/InMemoryEventStoreProjector.php
+++ b/src/Projection/InMemoryEventStoreProjector.php
@@ -15,7 +15,6 @@ namespace Prooph\EventStore\Projection;
 
 use ArrayIterator;
 use Closure;
-use Iterator;
 use Prooph\Common\Messaging\Message;
 use Prooph\EventStore\EventStore;
 use Prooph\EventStore\EventStoreDecorator;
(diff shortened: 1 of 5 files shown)


basz:
think this does the trick. eager to hear what you think. bas.

basz:
heads up. I identified a bug.

basz:
@prolic is there a reason you return *false* instead of *null* as defined by the interface?

https://github.com/prooph/pdo-event-store/blob/master/src/PdoStreamIterator.php#L188
https://php.net/manual/en/iterator.key.php

I would want to define `public function key(): ?int` within the MergedStreamIterator and simply return the contained (pdo) iterators key() result.

prolic (approved):

prolic:
I don't think the return value of `key()` is used anywhere, or is it?

basz:
> I don't think the return value of `key()` is used anywhere, or is it?

No it isn’t (i think)

PR #370, opened 2019-05-27
Allow marc-mabe/php-enum v4

Lctrs:


prolic:
Thanks
https://github.com/prooph/event-store/releases/tag/v7.5.4

Issue #371, opened 2019-06-03
Uncaught InvalidArgumentException: Could not create DateTimeImmutable


burzum:
I've asked about this problem some time ago in the Gitter chat and was told to update the dependency.  I'm using this `"prooph/event-store": "dev-master as 8.0.0-BETA-2",`. Well, I did so but I'm still getting this error for some reason.

All I'm doing to get this error is basically this:

```php
$events = $eventStore->readStreamEventsForward(TEST_UUID, 0, 2);
```

Resulting in this error.

```sh
PHP Fatal error:  Uncaught InvalidArgumentException: Could not create DateTimeImmutable from string "2019-06-03T09:15:34.5470932Z". in \vendor\prooph\event-store\src\Util\DateTime.php:45
Stack trace:
#0 \vendor\prooph\event-store-http-client\src\Internal\ResolvedEventParser.php(87): Prooph\EventStore\Util\DateTime::create('2019-06-03T09:1...')
#1 \vendor\prooph\event-store-http-client\src\Internal\EventStoreHttpConnection.php(428): Prooph\EventStoreHttpClient\Internal\ResolvedEventParser::parse(Array)
#2 \vendor\prooph\event-store-http-client\src\Internal\EventStoreHttpConnection.php(342): Prooph\EventStoreHttpClient\Internal\EventStoreHttpConnection->readStreamEventsForwardPolling('575c88f5-6763-4...', 0, 2, true, 0, NULL)
#3 \testdomain.php(141): Prooph\EventStoreHttpClient\Internal\EventStoreHttpConnection->readStreamEventsForward('575c88f5-6763-4...', 0, 2)
#4 {main in \vendor\prooph\event-store\src\Util\DateTime.php on line 45
```

As you can see the whole issue seems to come from the event-store library, not my code. We're using Event Store 4.1.1.0 (tested it also with Event Store 5.0.1.0) and the commit id of event-store is 0cbd53e4a521ab883daf45fe1fe67c435deeb8f7 The php version PHP 7.2.12 (cli) (built: Nov  8 2018 05:47:36) ( ZTS MSVC15 (Visual C++ 2017) x64 ).

The issue is that the format that is returned by the store or API doesn't seem to match what the store library expects: `2019-06-03T09:15:34.5470932Z` See the *seven* digits? [The documentation](https://www.php.net/manual/en/datetime.createfromformat.php) says for `u` up to *six* digits.

[...]

PR #374, opened 2019-06-14
Workaround for the inconsistent date format from the API

burzum:
See https://github.com/prooph/event-store/issues/371


prolic:
fixed in http client

PR #378, opened 2019-09-09
Fix incompatibility of Util\Assertion with beberlei/assert v3.2.3

codeliner:
Fixes:

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     "require": {
         "php": "^7.1",
         "marc-mabe/php-enum": "^2.3.1 || ^3.0.0 || ^4.0.0",
-        "prooph/common": "^4.1.0"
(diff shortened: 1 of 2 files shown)


codeliner:
Tried to fix the issue but we would need to pin beberlei/assert to 3.2.3. I'm not happy with such a solution. Opened an issue upstream and awaiting feedback: https://github.com/beberlei/assert/issues/287

prolic:
@codeliner let's wait for feedback on the assert lib issue and then go from there. We can postpone the next release a little.

codeliner:
upstream issue is fixed. I've added problematic versions of berberlei/assert to composer conflict section

prolic:
Great, can you take care of release? Thanks

Issue #366, opened 2019-04-11
Emit Enabled default configuration


enumag:
This is not exactly a bug per se but it certainly caused a wtf moment for me.

So this morning I noticed that my projection was not running. When I checked it's status, I found this:

```
  "status": "Faulted (Enabled)",
  "stateReason": "'emit' is not allowed by the projection/configuration/mode",
```

Then I digged what the emit option is, which was easy to find [here](https://eventstore.org/docs/projections/projections-config/index.html?tabs=tabid-5#emit-enabled) even with the error.

Then I checked how the projection was created. Prooph has a hard-coded `&emit=1` [here](https://github.com/prooph/event-store-http-client/blob/eea9e0e42d64f21fe4e7bd0070fa1d82a80a242e/src/Projections/ProjectionsManager.php#L192) which got me confused.

The problem was that I'm using `ProjectionsManager::updateQuery()` when migrating to a new version. This method however has the `$emitEnabled` option which I did not notice and it defaults to [false](https://github.com/prooph/event-store-http-client/blob/eea9e0e42d64f21fe4e7bd0070fa1d82a80a242e/src/Projections/ProjectionsManager.php#L436).

Maybe it's intended but it certainly caught me off-guard. When creating a projection there is no way to turn off the emit option, yet when updating it gets disabled by default.

Feel free to close this if this is intended. Just wanted to share my experience.

enumag:
ping @prolic

PR #382, opened 2019-11-22
Emit enabled inconsistency

prolic:
Resolves https://github.com/prooph/event-store/issues/366

### DIFF
--- a/src/Async/Projections/ProjectionsManager.php
+++ b/src/Async/Projections/ProjectionsManager.php
@@ -160,7 +160,7 @@ interface ProjectionsManager
     public function updateQueryAsync(
         string $name,
         string $query,
-        bool $emitEnabled = false,
+        ?bool $emitEnabled = null,
         ?UserCredentials $userCredentials = null
     ): Promise;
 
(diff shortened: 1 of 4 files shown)


src/Async/Projections/ProjectionsManager.php
```diff
@@ -160,7 +160,7 @@ public function getQueryAsync(string $name, ?UserCredentials $userCredentials =
     public function updateQueryAsync(
         string $name,
         string $query,
-        bool $emitEnabled = false,
+        bool $emitEnabled = null,
```
enumag:
```suggestion
        ?bool $emitEnabled = null,
```

Null means no change I'm guessing?

src/Projections/ProjectionsManager.php
```diff
@@ -144,7 +144,7 @@ public function getQuery(string $name, ?UserCredentials $userCredentials = null)
     public function updateQuery(
         string $name,
         string $query,
-        bool $emitEnabled = false,
+        bool $emitEnabled = null,
```
enumag:
```suggestion
        ?bool $emitEnabled = null,
```

src/Async/Projections/ProjectionsManager.php
prolic: yes

prolic:
Thanks for your review @enumag

PR #383, opened 2020-01-14
Fix ProjectionDetails

enumag:
Log report:
Screenshot from 2020-01-14 09-46-03


enumag:
Note: if we provide non-null defaults in https://github.com/prooph/event-store-client/pull/111 then this may not be necessary. I'm just unsure what the defaults should be.

prolic:
Thanks, good catch. I noticed some problems during some tests once, but couldn't reproduce quickly and didn't know where it came from.

Issue #389, opened 2020-04-15
Upcasting broken for projections in v7


marcuswinkler:
After upgrading from 7.5.1 to 7.5.6 I'm getting the following error:

>  [TypeError] Argument 2 passed to Prooph\EventStore\StreamIterator\MergedStreamIterator::__construct() must implement i
nterface Prooph\EventStore\StreamIterator\StreamIterator, instance of Prooph\EventStore\Upcasting\UpcastingIterator give
n, called in /var/www/vendor/prooph/event-store/src/Projection/InMemoryEventStoreReadModelProjector.php on line 331

The introduction of the `MergedStreamIterator` into the projectors in #367 seems to have broken upcasting. When the upcasting plugin is enabled, the event store returns an `UpcastingIterator`, which does not implement the `StreamIterator` interface required by `MergedStreamIterator`.

All the `StreamIterator` does over `Iterator` is adding the `Countable` interface. Is there a reason why `MergedStreamIterator` must be countable? Seems like an easy fix to just drop the `count()` method and make it work with simple iterators.

PR #391, opened 2020-04-17
fix upcasting iterator

prolic:
resolves https://github.com/prooph/event-store/issues/389

also:
- update copyright
- build php 7.3 + 7.4 on travis

ping @marcuswinkler can you test against this branch?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,18 @@ matrix:
     - php: 7.2
       env:
         - DEPENDENCIES="--prefer-lowest --prefer-stable"
+    - php: 7.3
+      env:
+        - DEPENDENCIES=""
+    - php: 7.3
+      env:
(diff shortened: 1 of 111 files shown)


prolic:
You can check the first and third commit for review only. The second one only updates the copyright and adds PHP 7.3 / 7.4 to travis.

marcuswinkler:
Works fine now in our application. This solution also avoids having to touch pdo-event-store.

I just want to note that the count method on the `UpcastingIterator` is not very reliable, as the `UpcastingIterator` can produce multiple events for every event on the inner iterator.

prolic:
Yes I know that, it's also mostly used for testing how far behind or up to
date a projection is, is a few events difference won't be a problem.

Thanks for reporting it.

PR #393, opened 2020-05-01
Make Guid class compatible with ramsey uuid 3+4

Nek-:
Problem: the composer.json was updated so it's possible to have
gedmo/uuid 4, besides the project is not compatible with it.

Solution: update Guid class.

### DIFF
--- a/src/Util/Guid.php
+++ b/src/Util/Guid.php
@@ -33,7 +33,9 @@ class Guid
 
     public static function generateAsHex(): string
     {
-        return self::generate()->getHex();
+        // string cast is a BC layer for ramsey/uuid v3
+        // to be refactor when dropping retrocompatibility
+        return (string) self::generate()->getHex();
     }
 
     public static function fromString(string $uuid): UuidInterface


src/Util/Guid.php
```diff
@@ -33,7 +33,9 @@ public static function generateString(): string
 
     public static function generateAsHex(): string
     {
-        return self::generate()->getHex();
+        // string cast is a BC layer for gedmo/uuid v3
```
enumag, line 36: don't you mean 
```suggestion
        // string cast is a BC layer for ramsey/uuid v3
```
?

src/Util/Guid.php
```diff
@@ -33,7 +33,9 @@ public static function generateString(): string
...
+        // to be refactor when dropping retrocompatibility
+        return (string) self::generate()->getHex();
```
prolic, line 38: why not simply `return self::generate()->getHex()->toString();` ?

src/Util/Guid.php
Nek-: Yeah sorry xD.

src/Util/Guid.php
Nek-: Well, that's actually more complicated lol.

But the why is because the point of this is to maintain BC with old v3 ramsey/uuid which returns a string (and `string->toString()` will obviously fail)

Nek-:
@prolic typo due to my weird brain.

prolic:
thanks @Nek-, I didn't catch that.

PR #395, opened 2020-05-02
mark dropSubscription on CatchUpSubscription as internal

prolic:


enumag:
I think I'd rather drop it from the interfaces and make it protected... I can have a look if the logic can be refactored in a way to make it private.

prolic:
Yeah, let's drop it.

PR #397, opened 2020-05-02
Remove dropSubscription method

enumag:


src/Async/EventStoreCatchUpSubscription.php
```diff
@@ -29,6 +28,4 @@ public function subscriptionName(): string;
     public function startAsync(): Promise;
```
prolic, line 28: what about this here? basically this is an implementation detail, how to start it and it should not get called by the end user.

src/EventStoreCatchUpSubscription.php
```diff
@@ -26,6 +26,4 @@ public function subscriptionName(): string;
     public function start(): void;
```
prolic, line 26: here we need to keep it anyway, because the user has to call this method initially.

src/Async/EventStoreCatchUpSubscription.php
enumag: Removed it both here and on the persistent subscription interface.

PR #398, opened 2020-05-03
Add Psalm

prolic:
remove EventAppearedOnSubscription and other similar interfaces, replace with closures

psalm error level 1 pass

see https://github.com/prooph/event-store-client/issues/127


prolic:
Almost done with psalm level 1, what left is mostly `MixedArgument` and `MixedArgumentTypeCoercion` in the `fromArray` methods of value objects. I don't know how to solve it in an elegant way, without specifying `@psalm-suppress` everywhere.

Can you have a look @enumag ?

prolic:
Also this:

```
ERROR: ImpurePropertyAssignment - src/EventId.php:41:9 - Cannot store a reference to an externally-mutable object inside an immutable object – consider using __clone (see https://psalm.dev/204)

        $this->uuid = $eventId;
```

It comes from ramsey uuid, how can I tell psalm this foreign lib is also immutable? Only `@psalm-suppress` again I suppose.

enumag:
Unfortunately I don't use Psalm in my projects yet so I can't answer your questions. Maybe if I try it myself.

jguittard:
@prolic You're consuming `3.8` version of `ramsey/uuid`
As you can see [here](https://github.com/ramsey/uuid/blob/3.x/src/UuidInterface.php), it's considered by default mutable.
Only `ramsey/uuid:^4` would have `UuidInterface` marked as immutable (see [here](https://github.com/ramsey/uuid/blob/master/src/UuidInterface.php#L27)

prolic:
@jguittard My question is not whether or not ramsey/uuid is marked as immutable or not. We know for a fact, that uuid's are immutable, so I like to tell psalm that it is. I don't maintain this uuid lib, so I cannot make changes to it. I assume using `@psalm-suppress` is the only way to solve it.
Unfortunately ramsey/uuid 4 is a broken library. Until they fix it, we can't allow version 4 in event-store.

prolic:
Don't merge yet, I let's wait until event-store-client is ready.

src/Async/EventStoreConnection.php
```diff
@@ -184,74 +186,127 @@ public function deletePersistentSubscriptionAsync(
...
     ): Promise;
 
     /**
+     * @param string $stream
+     * @param ?int $lastCheckpoint
+     * @param ?CatchUpSubscriptionSettings $settings
+     * @param Closure(EventStoreCatchUpSubscription, ResolvedEvent): Promise $eventAppeared
+     * @param ?Closure(EventStoreCatchUpSubscription): void $liveProcessingStarted
+     * @param ?Closure(EventStoreCatchUpSubscription, SubscriptionDropReason, ?Throwable): void $subscriptionDropped
+     * @param ?UserCredentials $userCredentials
```
enumag, line 211: Do PhpStorm, PHPStan and Psalm all understand `?` in annotations instead of `|null`?

enumag: By the way I heavily dislike duplicate annotations that do not add anything that's not already known from the typehints. So here I'd drop all the param annotations except the closures.

prolic: > Do PhpStorm, PHPStan and Psalm all understand `?` in annotations instead of `|null`?

phpstorm - no (maybe revert?)
psalm - yes
phpstan - i don't know

prolic: > By the way I heavily dislike duplicate annotations that do not add anything that's not already known from the typehints. So here I'd drop all the param annotations except the closures.

okay.

prolic:
I reverted `?Closure` to `null|Closure` in all docblocks.

src/UserManagement/UserDetails.php
```diff
@@ -17,79 +17,112 @@
...
+     * @param list<string> $groups
+     * @param list<RelLink> $links
+     */
+    private function __construct(
+        string $loginName,
+        string $fullName,
+        array $groups,
+        ?DateTimeImmutable $dateLastUpdated,
+        bool $disabled,
+        array $links)
```
enumag, line 42: newline before `)`

src/UserManagement/UserDetails.php
```diff
@@ -17,79 +17,112 @@
...
-        $details->loginName = $data['loginName'];
-        $details->fullName = $data['fullName'];
-        $details->groups = $data['groups'];
-        $details->disabled = $data['disabled'];
-
-        $details->dateLastUpdated = isset($data['dateLastUpdated'])
-            ? DateTime::create($data['dateLastUpdated'])
+        $dateLastUpdated = isset($data['dateLastUpdated'])
+            ? DateTime::create((string) $data['dateLastUpdated'])
             : null;
```
enumag, line 57: Use `??` operator.

prolic:
done

enumag:
I suggest merging this as is and increasing the error level in another PRs.

prolic:
@enumag this is already psalm level 1 and good to merge. If you review event-store-client (this is level 2 right now), we can merge both and work on psalm level 1 for the client implementationa afterwards.

PR #400, opened 2020-05-06
ProjectionManager can now return ValueObject when asking for status

### LINKED ISSUE #390: Let ProjectionsManager return objects instead of json-strings
I'm wondering about some methods of the ProjectionsManager.

These return ProjectionDetails objects: https://github.com/prooph/event-store/blob/master/src/Async/Projections/ProjectionsManager.php#L67-L86

This one however only returns json: https://github.com/prooph/event-store/blob/master/src/Async/Projections/ProjectionsManager.php#L88-L95

What's the reason? I have some cases where I'd like to get ProjectionDetails object for one specific projection...

unixslayer:
resolves #390

### DIFF
--- a/src/Async/EventStoreCatchUpSubscription.php
+++ b/src/Async/EventStoreCatchUpSubscription.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
 namespace Prooph\EventStore\Async;
 
 use Amp\Promise;
-use Throwable;
 
 interface EventStoreCatchUpSubscription
 {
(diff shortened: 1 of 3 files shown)


src/Projections/ProjectionStatus.php
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Projections;
+
+final class ProjectionStatus
```
enumag, line 16: Use the existing ProjectionDetails class. No need to duplicate it.

unixslayer: @enumag right, didn't notice that

Issue #401, opened 2020-05-08
Remove further json strings from managers


prolic:
f.e. `\Prooph\EventStoreClient\Projections\QueryManager::executeAsync`

there is more

/cc @unixslayer - interessted?

unixslayer:
@prolic ofc

prolic:
nice, can't wait for it :+1:

unixslayer:
@prolic this concerns both sync and async methods?

enumag:
@unixslayer yes

unixslayer:
@prolic sory for the delay. I think I finished this issue both in prooph/event-store and proopt/event-store-client and ready for review

PR #402, opened 2020-05-10
removing json strings

unixslayer:
resolves prooph/event-store#401

### DIFF
--- a/src/Projections/ProjectionStatistics.php
+++ b/src/Projections/ProjectionStatistics.php
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * This file is part of prooph/event-store.
+ * (c) 2014-2020 Alexander Miertsch <kontakt@codeliner.ws>
+ * (c) 2015-2020 Sascha-Oliver Prolic <saschaprolic@googlemail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Projections;
+
+final class ProjectionStatistics
+{
+    /** @var list<ProjectionDetails>  */
+    private array $projections;
+
+    /**
+     * @param list<ProjectionDetails> $projections
+     */
+    public function __construct(array $projections)
+    {
+        $this->projections = $projections;
+    }
+
+    /**
+     * @return list<ProjectionDetails>
+     */
+    public function projections(): array
+    {
+        return $this->projections;
+    }
+}
--- a/src/Projections/Query.php
+++ b/src/Projections/Query.php
@@ -1,0 +1,29 @@
+<?php
+
+/**
+ * This file is part of prooph/event-store.
+ * (c) 2014-2020 Alexander Miertsch <kontakt@codeliner.ws>
+ * (c) 2015-2020 Sascha-Oliver Prolic <saschaprolic@googlemail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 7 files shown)


src/Projections/ProjectionStatistics.php
```diff
@@ -0,0 +1,39 @@
...
+namespace Prooph\EventStore\Projections;
+
+use Webmozart\Assert\Assert;
+
+final class ProjectionStatistics
+{
+    private array $projections;
+
+    /**
+     * @param ProjectionDetails[]
```
prolic, line 23: `@param list<ProjectionDetails>`

src/Projections/ProjectionStatistics.php
```diff
@@ -0,0 +1,39 @@
...
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Projections;
+
+use Webmozart\Assert\Assert;
+
+final class ProjectionStatistics
+{
+    private array $projections;
```
prolic, line 20: `@var list<ProjectionDetails>`

src/Projections/ProjectionStatistics.php
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Projections;
+
+use Webmozart\Assert\Assert;
```
prolic, line 16: this is not a dependency

src/Projections/ProjectionStatistics.php
```diff
@@ -0,0 +1,39 @@
...
+final class ProjectionStatistics
+{
+    private array $projections;
+
+    /**
+     * @param ProjectionDetails[]
+     */
+    public function __construct(array $projections)
+    {
+        Assert::allIsInstanceOf($projections, ProjectionDetails::class);
```
prolic, line 27: not need, psalm will do static analysis

src/Projections/Query.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Prooph\EventStore\Projections;
+
+final class Query
```
prolic, line 16: Maybe rename to `QueryString`, @enumag ?

src/Projections/ProjectionStatistics.php
unixslayer: i think that `@param array<ProjectionDetails>` will be correct

src/Projections/ProjectionStatistics.php
unixslayer: right, didn't noticed that this is not direct dependency

src/Projections/ProjectionStatistics.php
unixslayer: This is not for static analysis, but for params validation, since array elements don't get runtime validation based on typehint. I understand that this is not necessary?

src/Projections/ProjectionStatistics.php
prolic: list<Foo> is a subtype of array<int, Foo>

src/Projections/ProjectionStatistics.php
prolic: Yes, we don't need this as we rely on psalm static analysis. It won't be checked on runtime.

src/Projections/Query.php
enumag: I'd keep Query personally.

PR #413, opened 2020-12-05
[7.x] Add PHP8 to travis / PHP8 support

fritz-gerneth:
- ~~Travis is currently missing the PHP 8 option ( https://travis-ci.community/t/php-8-0-missing/10132/17 ) but should be available next week~~
- This is mostly to see if/what needs to change to PHP8 support

Changes:
- [x] Updated the README badges to link to the correct branch for this version
- [x] Use `ProphesizeTrait`, `prophesize` integration has been [moved](https://github.com/sebastianbergmann/phpunit/issues/4141)  into its own package [`phpspec/prophecy-phpunit`](https://github.com/phpspec/prophecy-phpunit)
- [x] Replaced calls to `assertInternalType()` with new assertions ( old ones have been [removed](https://github.com/sebastianbergmann/phpunit/issues/3368) )
- [x] [`assertRegExp`](https://github.com/sebastianbergmann/phpunit/issues/4086) is deprecated and replaced 
- [x] Annotated a few tests with `@medium`  that were running longer than 1 second and thus being stopped
- [x] Update travis
- [x] Fix `at` deprecation warnings in phpunit
- [x] Update Doc-Header for 2021

PR #415, opened 2020-12-15
Fix issue 414, add tests + add php8 to travis (but not composer)

grymmare:
* Fixed bug caused by php 8 no longer returning false in some cases for `substr`
* Added tests for `DateTimeStringBugWorkaround`
* Added php 8 to travis as recommended by @enumag (but not composer)

*Please go gentle on my, this is my first PR in the open source community. Please let me know if I can improve anything. I'd like to continue to contribute 😄*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,10 @@ matrix:
     - DEPENDENCIES="--prefer-lowest --prefer-stable"
   - php: 8.0
     env:
-    - DEPENDENCIES="--ignore-platform-reqs"
+    - DEPENDENCIES="--ignore-platform-req=php"
   - php: 8.0
     env:
-    - DEPENDENCIES="--ignore-platform-reqs --prefer-lowest --prefer-stable"
+    - DEPENDENCIES="--ignore-platform-req=php --prefer-lowest --prefer-stable"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ matrix:
   - php: 7.4
     env:
     - DEPENDENCIES="--prefer-lowest --prefer-stable"
+  - php: 8.0
+    env:
+    - DEPENDENCIES="--ignore-platform-reqs"
+  - php: 8.0
+    env:
+    - DEPENDENCIES="--ignore-platform-reqs --prefer-lowest --prefer-stable"
 
 cache:
   directories:
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "php-coveralls/php-coveralls": "^2.2",
     "phpspec/prophecy": "^1.10.3",
     "phpunit/phpunit": "^9.1",
-    "prooph/php-cs-fixer-config": "^0.3.1",
+    "prooph/php-cs-fixer-config": "^0.4.0",
     "vimeo/psalm": "^3.11.2,<3.12"
   },
   "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
   ],
   "prefer-stable": true,
   "require": {
-    "php": "^7.4",
+    "php": "^7.4||^8.0",
     "ext-json": "*",
     "ramsey/uuid": "^3.9.3"
   },
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
   ],
   "prefer-stable": true,
   "require": {
-    "php": "^7.4||^8.0",
+    "php": "^7.4",
     "ext-json": "*",
     "ramsey/uuid": "^3.9.3"
   },
@@ -28,7 +28,7 @@
     "phpspec/prophecy": "^1.10.3",
     "phpunit/phpunit": "^9.1",
     "prooph/php-cs-fixer-config": "^0.3.1",
-    "vimeo/psalm": "^3.11.2"
+    "vimeo/psalm": "^3.11.2,<3.12"
   },
   "autoload": {
     "psr-4": {
--- a/src/Internal/DateTimeStringBugWorkaround.php
+++ b/src/Internal/DateTimeStringBugWorkaround.php
@@ -23,7 +23,6 @@ final class DateTimeStringBugWorkaround
     {
         $micros = \substr($dateTimeString, 20, -1);
 
-        // PHP versions before v8 returned false instead of an empty string
         if (false === $micros || '' === $micros) {
             // no microseconds given
             $dateTimeString = \substr($dateTimeString, 0, 19) . '.';
--- a/src/Internal/DateTimeStringBugWorkaround.php
+++ b/src/Internal/DateTimeStringBugWorkaround.php
@@ -23,7 +23,8 @@ final class DateTimeStringBugWorkaround
     {
         $micros = \substr($dateTimeString, 20, -1);
 
-        if (false === $micros) {
+        // PHP versions before v8 returned false instead of an empty string
+        if (false === $micros || '' === $micros) {
             // no microseconds given
             $dateTimeString = \substr($dateTimeString, 0, 19) . '.';
             $micros = '';
--- a/tests/Internal/DateTimeStringBugWorkaroundTest.php
+++ b/tests/Internal/DateTimeStringBugWorkaroundTest.php
@@ -1,8 +1,17 @@
 <?php
 
+/**
+ * This file is part of prooph/event-store.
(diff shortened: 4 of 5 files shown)


src/Internal/DateTimeStringBugWorkaround.php
```diff
@@ -23,7 +23,8 @@ public static function fixDateTimeString(string $dateTimeString): string
     {
         $micros = \substr($dateTimeString, 20, -1);
 
-        if (false === $micros) {
+        // PHP versions before v8 returned false instead of an empty string
```
enumag, line 26: The comment is not necessary in my opinion.

```suggestion
```

tests/Internal/DateTimeStringBugWorkaroundTest.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public function test_it_correctly_converts_datetime_with_less_than_6_micros()
+    {
+        $date = "2020-12-14T12:55:57.1234Z";
+        $expectedConvertedDate = "2020-12-14T12:55:57.123400Z";
+        $convertedDate = DateTimeStringBugWorkaround::fixDateTimeString($date);
+        $this->assertEquals($convertedDate, $expectedConvertedDate);
+    }
+}
```
enumag, line 35: Add a newline at the end of a file.

tests/Internal/DateTimeStringBugWorkaroundTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace ProophTest\EventStore\Internal;
+
+
+use PHPUnit\Framework\TestCase;
+use Prooph\EventStore\Internal\DateTimeStringBugWorkaround;
+
+class DateTimeStringBugWorkaroundTest extends TestCase
+{
+    public function test_it_correctly_converts_datetime_with_missing_micros()
```
enumag, line 12: `: void`

tests/Internal/DateTimeStringBugWorkaroundTest.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $date = "2020-12-14T12:55:57Z";
+        $expectedConvertedDate = "2020-12-14T12:55:57.000000Z";
+        $convertedDate = DateTimeStringBugWorkaround::fixDateTimeString($date);
+        $this->assertEquals($convertedDate, $expectedConvertedDate);
```
enumag, line 17: iirc expected value is supposed to be the first argument and the actual value the second one, you swapped them

tests/Internal/DateTimeStringBugWorkaroundTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+
+use PHPUnit\Framework\TestCase;
+use Prooph\EventStore\Internal\DateTimeStringBugWorkaround;
+
+class DateTimeStringBugWorkaroundTest extends TestCase
+{
+    public function test_it_correctly_converts_datetime_with_missing_micros()
+    {
+        $date = "2020-12-14T12:55:57Z";
```
enumag, line 14: Use apostrophes instead of quotes for strings please.

tests/Internal/DateTimeStringBugWorkaroundTest.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+
+namespace ProophTest\EventStore\Internal;
+
+
+use PHPUnit\Framework\TestCase;
+use Prooph\EventStore\Internal\DateTimeStringBugWorkaround;
+
+class DateTimeStringBugWorkaroundTest extends TestCase
```
enumag, line 10: In this case I think you should write just one test with a `@dataProvider`, not separate methods.

enumag (changes requested):

enumag:
> Please go gentle on my, this is my first PR in the open source community. Please let me know if I can improve anything. I'd like to continue to contribute 😄

@grymmare Thanks for your first opensource contribution! I added a code review for minor improvements. ;-) No worries about these minor mistakes, happens to everyone the first time. ^^

fritz-gerneth:
Not sure why Travis is not running on this PR. I don't think with the current setup of dependencies this is actually installable on PHP8 at the moment. See #413 for the initial discussion & errors.

tests/Internal/DateTimeStringBugWorkaroundTest.php
grymmare: Ah, I was not aware of this feature. Rewrite the test and addressed your  other comments.

grymmare:
> Not sure why Travis is not running on this PR. I don't think with the current setup of dependencies this is actually installable on PHP8 at the moment. See #413 for the initial discussion & errors.

It seems to have been run now. With errors though, but they are psalm related (pure function returning member not marked as pure)

fritz-gerneth:
I only noticed now this is targetting the master branch (v8) and not the 7.x branch :)

.travis.yml
```diff
@@ -11,6 +11,12 @@ matrix:
   - php: 7.4
     env:
     - DEPENDENCIES="--prefer-lowest --prefer-stable"
+  - php: 8.0
+    env:
+    - DEPENDENCIES="--ignore-platform-req=php"
+  - php: 8.0
+    env:
+    - DEPENDENCIES="--ignore-platform-req=php --prefer-lowest --prefer-stable"
```
fritz-gerneth, line 19: I don't think we should ignore these here. Better fix these issues properly. Not doing so makes this either not installable for normal users or it forces them to use this flag on their project too.

grymmare: Sure, makes sense, but without the flag, we cannot run the tests for php 8 until all dependencies supports php 8. What if we don't add php8 to composer.json, that way we don't officially support php8, but can still run tests. 

The tests that fail (in addition to dependencies not supporting php8 yet) are related to the vimeo/psalm version. 
* Psalm 3.11.2 is successful
* Psalm 3.18.2 fails. Seems that marking a class with @psalm-immutable conflicts with @psalm-pure somehow. Maybe this can be resolved by specifying a version constraint, but I guess this depends on how psalm develops, or removing all @psalm-pure annotations in classes annotated with @psalm-immutable

prolic:
As php-cs-fixer can be installed with PHP 8, but it's not possible to run yet, we need to disable these checks on travis until it's supported to check on PHP 8 as well.

grymmare:
> As php-cs-fixer can be installed with PHP 8, but it's not possible to run yet, we need to disable these checks on travis until it's supported to check on PHP 8 as well.

But, doesn't this only apply if we would start using php 8 features, which we cannot if we want to support 7.4?

---

The tests that fail are the tests that uses the latest dependencies, so if they were successful before, probably they fail now due to changes in psalm and php-cs-fixer, and so they will fail in any new commit. 
image

Mostly psalm related and fails for both PHP 7.4 and PHP 8.0 (not related to this PR, but has to be fixed)
image

The php-cs-fixer failures is minor (easy to fix and is in my added test file)
image




@prolic what do you think?

composer.json
```diff
@@ -28,7 +28,7 @@
     "phpspec/prophecy": "^1.10.3",
     "phpunit/phpunit": "^9.1",
     "prooph/php-cs-fixer-config": "^0.3.1",
-    "vimeo/psalm": "^3.11.2"
+    "vimeo/psalm": "^3.11.2,<3.12"
```
prolic, line 31: why this `<3.12` ?

prolic:
see https://github.com/prooph/php-cs-fixer-config/releases/tag/v0.4.0

composer.json
grymmare: Because later versions adds psalm warnings which fails the build. (see log here: [https://travis-ci.com/github/prooph/event-store/jobs/461215042](https://travis-ci.com/github/prooph/event-store/jobs/461215042)) 

My suggestion is to fix this in another PR. If you disagree though, we need to make the psalm checks successful. I havn't found any other solution than removing all the affected @psalm-pure annotations.

grymmare:
> see https://github.com/prooph/php-cs-fixer-config/releases/tag/v0.4.0

Thanks. Updated to use this version

composer.json
grymmare: Found this: https://github.com/vimeo/psalm/issues/4114

Instead of @psalm-pure, use @psalm-mutation-free 

I can change this throughout the code (more than 100 places I think) and fix some other psalm related stuff. Just let me know.

prolic:
@grymmare sorry I didn't check this earlier. My bad.

PR #426, opened 2022-05-10
PHP 8.1, amp v3

prolic:
- Require PHP 8.1
- make use of Fibers
- require amp v3


enumag:
I'm sorry @prolic but this is simply not reviewable. You're mixing copyright changes, update to 8.1 enums, constant renames etc. along with the Fibers update. Way too much noise. Each of these changes should be a separate commit.

prolic:
Psalm requires amp v2. Too bad. Maybe https://github.com/bamarni/composer-bin-plugin can help or just working with psalm from phar file.

For the rest, you are right, but the fastest way to get this huge refactoring done, was to simply do all changes and afterwards adjust the test-suite once.

Sorry if this makes reviewing a lot harder, but my time is also very limited.

PR #427, opened 2022-11-04
Php81 dependencies comp

basz:


basz:
it passes now.

I should only update the readme. php73 was dropped. under what version would this be released

PR #430, opened 2023-01-06
PHP 8.2 compatibility (and some more)

zluiten:
Most of these changes are done with the help of [rector](https://github.com/rectorphp/rector).

  * Ensure compatibility with PHP 8.2 (dynamic properties)
  * Add explicit public visibility to remaining constants
  * Use array spread operator instead of array_merge where possible
  * Add null default to nullable properties properties that have no default
  * Add typed properties and return types where possible
  * Simplify empty() function calls on empty arrays
  * Update licenses to 2023

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Prooph Event Store
 
-PHP 7.3 EventStore Implementation.
+PHP 7.4 EventStore Implementation.
 
 [![Build Status](https://travis-ci.com/prooph/event-store.svg?branch=7.x)](https://travis-ci.com/prooph/event-store)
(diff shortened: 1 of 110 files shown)


fritz-gerneth:
Not having checked every single file, but looks good to me

prolic (approved):

prolic:
Thanks @zluiten! Great work!

prolic:
Currently some builds are failing, @zluiten if you have time to look into those, would be great. Otherwise I try to check when I have time. I'd like to have all green builds before a release.

zluiten:
Thanks @prolic for the quick review/merge! The fix is here #431.

szepczynski:
is any chance to support php 8?

PR #442, opened 2025-03-27
Allow to configure MetadataMatcher separately for ReadModelProjector

unixslayer:
Allow to configure MetadataMatcher separately for ReadModelProjector. 

Certain metadata can be global, shared between streams.

### DIFF
--- a/src/Projection/InMemoryEventStoreReadModelProjector.php
+++ b/src/Projection/InMemoryEventStoreReadModelProjector.php
@@ -131,14 +131,25 @@ final class InMemoryEventStoreReadModelProjector implements ReadModelProjector
         return $this;
     }
 
-    public function fromStream(string $streamName, ?MetadataMatcher $metadataMatcher = null): ReadModelProjector
+    public function withMetadataMatcher(?MetadataMatcher $metadataMatcher = null): ReadModelProjector
+    {
+        $this->metadataMatcher = $metadataMatcher;
+
+        return $this;
+    }
+
+    public function fromStream(string $streamName/**, ?MetadataMatcher $metadataMatcher = null*/): ReadModelProjector
     {
         if (null !== $this->query) {
             throw new Exception\RuntimeException('From was already called');
         }
 
+        if (\func_num_args() > 1) {
+            \trigger_error('The $metadataMatcher parameter is deprecated. Use withMetadataMatcher() instead.', E_USER_DEPRECATED);
+            $this->metadataMatcher = \func_get_arg(1);
+        }
+
         $this->query['streams'][] = $streamName;
-        $this->metadataMatcher = $metadataMatcher;
 
         return $this;
     }
--- a/src/Projection/ReadModelProjector.php
+++ b/src/Projection/ReadModelProjector.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Prooph\EventStore\Projection;
 
 use Closure;
+use Prooph\EventStore\Metadata\MetadataMatcher;
 
 interface ReadModelProjector
 {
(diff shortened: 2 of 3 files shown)


src/Projection/InMemoryEventStoreReadModelProjector.php
```diff
@@ -131,14 +131,24 @@ public function init(Closure $callback): ReadModelProjector
...
+    }
+
+    public function fromStream(string $streamName/**, ?MetadataMatcher $metadataMatcher = null*/): ReadModelProjector
     {
         if (null !== $this->query) {
             throw new Exception\RuntimeException('From was already called');
         }
 
+        if (\func_num_args() > 1) {
+            $this->metadataMatcher = \func_get_arg(1);
```
basz, line 149: maybe keep the argument and add deprecation?

```
    /**
     * @param string $streamName The name of the stream.
     * @param MetadataMatcher|null $metadataMatcher [DEPRECATED] Use withMetadataMatcher() instead.
     * @return ReadModelProjector
     * @deprecated The $metadataMatcher parameter is deprecated. Use withMetadataMatcher() instead.
     */
    public function fromStream(string $streamName, ?MetadataMatcher $metadataMatcher = null): ReadModelProjector
    {
        if ($metadataMatcher !== null) {
            trigger_error(
                'The $metadataMatcher parameter is deprecated. Use withMetadataMatcher() instead.',
                E_USER_DEPRECATED
            );
        }
        ...
```

basz (changes requested):

src/Projection/InMemoryEventStoreReadModelProjector.php
unixslayer: @basz `InMemoryEventStoreReadModelProjector::fromStream()` is anyway incompatible with `ReadModelProjector` interface due `$metadataMatcher` is not defined in there. Although, I've added `trigger_error`. This will give clear information about deprecation if someone use that parameter to switch for `withMetadataMatcher`.

unixslayer:
@basz I've also opened similar PR for PDO EventStore https://github.com/prooph/pdo-event-store/pull/250

unixslayer:
@basz when can this be released?

basz:
done. thanks @unixslayer

zluiten:
Nice addition, but adding an interface method is actually a bc break.

basz:
Shite. you are right.

What the way out of this hole? Probably a new patch release undoing this addition. Redoing it in a major.

Problem is an 8.x branch already exists and I have no idea what the status is of that.

@prolic @zluiten suggestions about the 8.x branch?

zluiten:
One way is by adding a separate interface that extends the existing one. This way, projects that have their implementation can choose to migrate. Doing it like this can still be done in a minor or patch release.

```php
interface MetadataAwareReadModelProjector extends ReadModelProjector 
{
    public function withMetadataMatcher(?MetadataMatcher $metadataMatcher = null): ReadModelProjector;
}
```

[prooph/pdo-eventstore](https://github.com/prooph/pdo-event-store/) could probably implement this in a minor/patch release as the PdoEventStoreReadModelProjector is final so it cannot be extended.

prolic:
Adding a separate interface extending the existing one is the way to go
here.

For 8.x branch: it's actually ready for final release, I just never had the
time to review the docs and make sure they are in sync.

unixslayer:
I'll deal with this BC Break. Sorry for the inconvenience. I realized today myself what I've done.

basz:
no worries. i could have catched that in the review.

@prolic I won't be touching 8.x as i have no idea what the differ functionally...

PR #443, opened 2025-03-31
PHP 8.1 is still supported

unixslayer:
@basz @prolic why PHP 8.1 was dropped as it is still supported?

### DIFF
--- a/src/Projection/InMemoryEventStoreReadModelProjector.php
+++ b/src/Projection/InMemoryEventStoreReadModelProjector.php
@@ -131,14 +131,25 @@ final class InMemoryEventStoreReadModelProjector implements ReadModelProjector
         return $this;
     }
 
-    public function fromStream(string $streamName, ?MetadataMatcher $metadataMatcher = null): ReadModelProjector
+    public function withMetadataMatcher(?MetadataMatcher $metadataMatcher = null): ReadModelProjector
+    {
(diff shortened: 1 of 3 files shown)


basz:
It was decided to drop after a short discussion.

Due to https://github.com/prooph/common/pull/85#issuecomment-2647986095

We *should* update all mentions of that...

unixslayer:
@basz phpunit:9 was installed due prooph/php-cs-fixer-config had support for it. It would be sufficient to drop it there which let to keep support for PHP 8.1. I've already opened PR for that https://github.com/prooph/php-cs-fixer-config/pull/12 

Are you willing to bring back support for PHP 8.1? I can take prepare other PRs.

basz:
o well that would be great i guess. if you you still need support for it, can get it to work with 8.1 and do the work. I'm all for it.

q: do you still use 8.1?

unixslayer:
@basz yes, I have one project that still runs on 8.1 and won't be upgraded until a few weeks before it loses the support due to some internal politics. i can imagine others in a similar situation and am surprised no one raised that yet.

This also brought a BC break for Ecotone library that uses Prooph's Event Store and still supports 8.1.

unixslayer:
https://github.com/prooph/php-cs-fixer-config/pull/12 here is the first PR

PR #446, opened 2025-04-01
PHP 8.1 is still supported

unixslayer:


basz:
https://github.com/prooph/php-cs-fixer-config/releases/tag/v0.7.0 has been released, so you could update the dependency and it should pass...

unixslayer:
@basz dependency updated

PR #448, opened 2025-04-03
adds roave bc checker

basz:
test this

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -35,7 +35,8 @@ jobs:
         with:
           fetch-depth: 0
       - name: "Check for BC breaks"
-        uses: docker://nyholm/roave-bc-check-ga
+        run: |
+          docker run -u $(id -u) -v $(pwd):/app nyholm/roave-bc-check-ga
 
   tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


basz:
weird. locally these are filtered out...

https://github.com/prooph/event-store/actions/runs/14241242016/job/39911435244?pr=448#step:3:700

basz:
@unixslayer what do you think?

unixslayer:
@basz seems like it checks for BC Breaks in vendor libraries instead of making sure we are not introducing any. Maybe ask @Ocramius?

Ocramius:
It checks for BC breaks you are introducing in your library, exposed to others.

Vendor libraries changing can lead to changes in your libraries, which in turn can lead to exposed BC breaks for others, but that's just one possible case

.github/workflows/integration.yaml
```diff
@@ -28,6 +28,17 @@ jobs:
...
 
+  bc-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: "Check for BC breaks"
+        run: |
+          composer require --dev roave/backward-compatibility-check
```
Ocramius: Consider pinning to an exact version.

Ideally, you would lock with a dedicated `composer.lock`

basz:
Don't think it does that. I've tested locally. Added a method to an interface and committed. Then ran the check and it blocked.

unixslayer:
> Don't think it does that. I've tested locally. Added a method to an interface and committed. Then ran the check and it blocked.

ok. I didn't commit when testing locally but when I did, it showed me I'm introducing a BC Break. Maybe only pin to an exact version like @Ocramius suggested and we're good to go.

PR #449, opened 2025-04-18
PHPUnit compatibility

unixslayer:
- upgrade to PHPUnit 10
- remove enforcing time limit as it may mark tests as risky
- test classes should be suffixed with `Test`
- abstract Test classes should be suffixed with `TestCase`
- remove remaining deprecations

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -60,7 +60,7 @@ jobs:
           - php: 8.4
             dependencies: ''
           - php: 8.4
-            depdendencies: '--prefer-lowest --prefer-stable'
+            dependencies: '--prefer-lowest --prefer-stable'
 
     steps:
       - uses: actions/checkout@v2
--- a/composer.json
+++ b/composer.json
@@ -29,15 +29,14 @@
         "prooph/common": "^4.6.0"
     },
     "require-dev": {
+        "php-coveralls/php-coveralls": "^2.7",
         "phpspec/prophecy": "^1.10.3",
         "phpspec/prophecy-phpunit": "^2.0",
-        "phpunit/php-invoker": "^3.1",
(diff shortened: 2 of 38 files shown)


unixslayer:
CC @prolic @basz

basz:
PHPUnit does not run abstract test classes so their naming does not matter in the context of PhpUnit. Test classes should end with Test by default. So to prevent a BC for consuming libraries (pdo-event-store) I believe we can keep the naming for the abstract classes as-is.

unixslayer:
> PHPUnit does not run abstract test classes so their naming does not matter in the context of PhpUnit. Test classes should end with Test by default. So to prevent a BC for consuming libraries (pdo-event-store) I believe we can keep the naming for the abstract classes as-is.

@basz PHPUnit does not run abstract test classes but their naming matters as starting from version 10 abstract test classes should be suffixed with `TestCase` resulting with a warning if not.

I noticed, some of the abstract test classes are extended in `prooph/pdo-event-store`. I will anyway continue this topic there in https://github.com/prooph/pdo-event-store/pull/258 so eventual BC Breaks will be solved right after this PR.

unixslayer:
also, all bc-check fails seems to be caused by upgrading PHPUnit so I'd consider that as a false-positive

basz:
i want to release as 7.12.3 with a note

*Note: any abstract test classes have been renamed due to phpunit changes. Any tests depending on one of those should refactor*

Think that's enough?

unixslayer:
> i want to release as 7.12.3 with a note
> 
> _Note: any abstract test classes have been renamed due to phpunit changes. Any tests depending on one of those should refactor_
> 
> Think that's enough?

@basz should be enough as tests should be considered as private. if someone extends them (beside `prooph/pdo-event-store`) they should be aware of possible and potential BC Breaks.

PR #450, opened 2025-12-22
Improve return types to prevent deprecation warnings

zluiten:
## What changed

- `UpcastingIterator::key()` gets a native `mixed` return type, replacing the `#[\ReturnTypeWillChange]` attribute and `@return mixed` docblock.
- `MergedStreamIterator::current()` and `UpcastingIterator::current()` retain their existing signatures - see below for why.
- `prooph/bookdown-template` is removed from `require-dev` as it requires `league/commonmark ~0.7` which only resolves to `0.19.x-dev`, incompatible with PHP 8+. This was breaking the entire CI pipeline.

----

`UpcastingIterator::key()` is safe because `UpcastingIterator` is `final` - no subclasses can exist, so adding a return type cannot break anyone.

The other two cannot use native types without a BC break:

- **`MergedStreamIterator::current()`** - `MergedStreamIterator` is not `final`. Adding any return type to a previously untyped method forces subclasses to declare a compatible type or receive a deprecation warning.
- **`UpcastingIterator::current()`** - already has `?Message`. Widening it to `mixed` is flagged as non-covariant by the BC checker, as callers may rely on the narrower type. Since `?Message` already satisfies `Iterator::current(): mixed` covariantly, no deprecation warning is generated and no change is needed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,6 @@
         "phpspec/prophecy": "^1.10.3",
         "phpspec/prophecy-phpunit": "^2.0",
         "phpunit/phpunit": "^10.0",
-        "prooph/bookdown-template": "^0.2.3",
         "prooph/php-cs-fixer-config": "^0.7.0",
         "psr/container": "^1.0",
         "sandrokeil/interop-config": "^2.0.1"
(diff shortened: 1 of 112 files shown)


zluiten:
@basz Can you have look?

basz (approved):

